PR #34, opened 2015-03-11
Unmarshal arbitrary length DH generators

s-zanella:


dadrian:
Is there a reason for doing this? I explicitly didn't implement generators using bigINts because the serialization of a bigInt to json is awful and I have yet to see a generator larger than one byte.

s-zanella:
I think that configurations with larger generators are common. 
One way of generating DH parameters with a large generator is using OpenSSL dhparam with the -dsaparam option:

	$ openssl dhparam -dsaparam 512 -text
	...
	Diffie-Hellman-Parameters: (512 bit)
	    prime:
	        00:ce:99:7e:ad:5a:44:86:9c:4d:aa:86:ac:2e:26:
	        75:e7:a4:65:8d:e0:d4:54:8d:48:86:17:18:09:17:
	        2a:2d:9c:85:eb:7d:d0:02:7e:dd:e9:f0:4e:79:1b:
	        61:bc:d4:72:e5:78:de:90:44:da:75:ad:6e:cf:65:
	        f4:0e:a8:7d:b9
	    generator:
	        6c:c2:41:86:b1:fc:04:ad:53:62:46:95:ab:a0:2e:
	        9c:41:60:3b:66:56:aa:c4:17:21:b0:5b:a3:5f:cd:
	        71:4b:4b:d1:d2:ef:ca:85:a1:1d:cb:90:07:e1:d7:
	        4d:df:19:40:8e:a8:6c:98:97:f3:cc:ec:00:15:47:
	        7c:5f:31:cd

PR #38, opened 2015-04-03
Use wait groups instead of pipes

corny:
It is more convenient in this place.


dadrian:
Thanks!

dadrian:
That's actually a much nicer implementation, I didn't realize WaitGroup existed when I originally implemented this.

PR #39, opened 2015-04-04
Extract String() method from MarshalJSON

corny:
This makes it possible to retrieve the name without marshalling the struct to JSON.
I need it because I use zgrab with my own worker.

PR #40, opened 2015-04-04
Revert "Extract String() method from MarshalJSON"

dadrian:
Reverts zmap/zgrab#39. 
Previous commit was resulting in corrupted names. This is what I get for testing after merging.


corny:
I am going to write some tests.

corny:
Solved in #41

PR #41, opened 2015-04-04
Extract String() method from MarshalJSON [fixed]

corny:
This makes it possible to retrieve the name without marshalling the struct to JSON.

`x.String()` works better than `string(x)`


dadrian:
It looks like this PR contains some of the CertificateFingerprint changes, can you rebase it on top of latest master?

corny:
Done

dadrian:
Thanks!

corny:
I also have to say thank you for this great piece of software!
I am writing on my thesis and use it to do research on TLS certificates of mail servers.

PR #42, opened 2015-04-05
Gracefully handle negative serial numbers

corny:
According to https://tools.ietf.org/html/rfc5280#section-4.1.2.2
```
Non-conforming CAs may issue certificates with serial numbers
that are negative or zero.  Certificate users SHOULD be prepared to
gracefully handle such certificates.
```
This issue is also addressed by https://github.com/golang/go/issues/8265

From 208314 hosts 662 were affected by this problem. Some examples:
* 93.57.50.24
* 123.100.249.85
* 202.43.34.79


dadrian:
I'm all for making our certificate parsing more lenient, but do we know how this change interacts with the rest of the library?

corny:
At least for me it works after this change.
There are only few lines of code that deal with the `SerialNumber`.

PR #44, opened 2015-04-06
Refactor fingerprint creation

corny:


dadrian:
Is there a reason the slices passed by pointer to the Fingerprint() functions?

corny:
Is `[]byte(in.Raw)` really a slice? I want to avoid call by value and I am not sure if I need pointers in this case.

dadrian:
`in.raw` is already a byte slice, so you should just be able to pass it directly without even calling `raw := []byte(in.Raw)`. 

Also, all of the hash libraries still need to be imported for side-effects so that `crypto.RegisterHash` is called for each hash function used during the verification process, e.g.

```
import _ "crypto/md5"
```

corny:
I updated my pull request. I also:
* renamed String() to Hex()
* added SHA512Fingerprint() for future use

dadrian:
Thanks!

PR #46, opened 2015-04-09
Provide more validation errors

corny:
This pull request is not intended to be merged (yet).

A certificate might be invalid for several reaons. I think it is useful to get to know them all.
My pull request modifies the validation methods to return multiple errors instead of only the first one.
It would also be nice to get messages that are easier to parse. How could we achieve that?

### DIFF
--- a/ztools/x509/json.go
+++ b/ztools/x509/json.go
@@ -89,11 +89,11 @@ type jsonTBSCertificate struct {
 }
 
 type jsonSignature struct {
-	Value           []byte `json:"value"`
-	Valid           bool   `json:"valid"`
-	ValidationError string `json:"validation_error,omitempty"`
-	Matches         *bool  `json:"matches_domain"`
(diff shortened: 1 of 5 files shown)


dadrian:
I think it probably makes more sense to encapsulate all the errors in a single custom error struct. This object may contain a `[]string`. Then we define an encoding of that error object to JSON. In this way, functions that call validate can stay the same, but the error will be more informative when it's encoded.

zakird:
What's the status of this pull request? It looks like it's been open for nearly a year now.

dadrian:
Our validation is better now than it was; and we're able to discern between "some error" and "name mismatch" correctly now. The `validation_error` string is reasonably descriptive, and will contain the first (non-name) error encountered.

zakird:
I'm going to close this at this point then. We can reopen if needed.

corny:
It's outdated. Thanks for closing.

PR #47, opened 2015-04-09
Backport x509 changes from the Go repository

corny:
I copied all files from the go/x509 directory to ztools/x509 and commited changes are not specific to zgrab.


ztools/x509/cert_pool.go
```diff
@@ -77,14 +77,13 @@ func (s *CertPool) AddCert(cert *Certificate) {
 }
 
 // AppendCertsFromPEM attempts to parse a series of PEM encoded certificates.
-// It appends any certificates found to s and returns true if ALL certificates
```
dadrian: We actually made this change ourselves, to make the function be more strict when loading in a Root CA file.

corny: Ok, so only the `cert_pool.go` need to keep untouched?

dadrian: If there were other changes, that's fine. I just want to make sure this function keeps our modified semantics.

corny: I reverted my changes of this function and added a `added by zgrab` notice to line.

corny:
Has anything to be done before you can merge this pull request?

dadrian:
Sorry, I'm just running behind on this. I'll try to get this merged later today.

corny:
Don't hurry - I just wanted to know if there is anything I should do.

dadrian:
Not off the top of my head. I just need to sit down and make sure the changes don't interact with anything we do in an unexpected way.

dadrian:
I started going through this a bit more. First off, we should make sure not to commit any files that are created automatically by go generate (e.g. `root_darwin_arm.go`).

Also, what git revision or go release did you pull these changes from?

corny:
I used the latest version from: https://github.com/golang/go/blob/88399b2e4662ad5a8ca8dc50c5d60fbe42d86396/src/crypto/x509/x509.go

As far as I understand the `root_darwin_arm_gen.go` it is a simple program that generates the `root_darwin_arm.go`. They are both in the go repository: https://github.com/golang/go/tree/88399b2e4662ad5a8ca8dc50c5d60fbe42d86396/src/crypto/x509

PR #53, opened 2015-05-12
Backport crypto/x509 and crypt/tls changes

corny:
I need SHA384 for my project and I want to respect RFC 7525. It was too hard to make seperate commits. I will test it extensively and I hope my changes do not brake anything.

- Merge https://github.com/golang/go/tree/1ddb8c20c6f686170c506c6440e194a58c040bec from the Go repository
- Ciphersuite order is chosen by [RFC 7525](https://tools.ietf.org/html/rfc7525)
- Move cipher suite lists to ztls/cipher_suite_lists.go

Important new features in Go 1.5:
- Support for SHA384 cipher suites
- Allow parsing of certificates with unknown critical extensions

It requires the current development version of Go 1.5 (not released yet).


dadrian:
So now that 1.5 has been released, we should probably merge the actual changes back in. Beyond that, we've made other changes to the X.509 lib, and these backported changes likely don't match.

I'm going to close this for now and create an issue to track this moving forward.

corny:
:+1:

PR #67, opened 2015-09-05
Add Bytes() to CipherSuite and TLSVersion

corny:
I need these to get the version and cipher suite as byte array in my application.


dadrian:
Thanks!

PR #69, opened 2015-10-20
Zzma/follow redirects

zzma:
@dadrian @zakird 

Basic HTTP / HTTPS redirects under the following conditions:
1) the --http-max-redirects option is set > 0
2) domain is specified alongside/after IP address (separated by a comma)
3) the redirect location matches the domain, and has the same HTTP or HTTPS protocol as the original connection
4) the initial HTTP response has the "Keep-Alive" header or HTTP 1.1, which uses persistent connections by default

What is the best way to test this? I have only verified that it works with a local test Rails server. I'm hesitant to try the Alexa Top Million from my laptop...

### DIFF
--- a/zlib/conn.go
+++ b/zlib/conn.go
@@ -32,7 +32,9 @@ import (
 	"github.com/zmap/zgrab/ztools/ssh"
 	"github.com/zmap/zgrab/ztools/util"
 	"github.com/zmap/zgrab/ztools/x509"
+	"github.com/zmap/zgrab/ztools/zlog"
 	"github.com/zmap/zgrab/ztools/ztls"
+	"io"
 )
 
 var smtpEndRegex = regexp.MustCompile(`(?:^\d\d\d\s.*\r\n$)|(?:^\d\d\d-[\s\S]*\r\n\d\d\d\s.*\r\n$)`)
@@ -352,6 +354,9 @@ func (c *Conn) doHTTP(config *HTTPConfig) error {
 			}
 
 			if httpResponse, err = c.sendHTTPRequestReadHTTPResponse(redirectBaseRequest, config); err != nil {
+				if err == io.ErrUnexpectedEOF {
+					zlog.Errorf("Connection closed before making redirect to %s (%s)", c.domain, c.RemoteAddr())
+				}
 				return err
 			}
 
--- a/zlib/conn.go
+++ b/zlib/conn.go
@@ -48,20 +48,20 @@ const (
 // Implements the net.Conn interface
 type Conn struct {
 	// Underlying network connection
-	conn    net.Conn
-	tlsConn *ztls.Conn
-	isTls   bool
+	conn                net.Conn
+	tlsConn             *ztls.Conn
+	isTls               bool
 
-	grabData GrabData
+	grabData            GrabData
 
 	// Max TLS version
-	maxTlsVersion uint16
+	maxTlsVersion       uint16
 
 	// Cache the deadlines so we can reapply after TLS handshake
-	readDeadline  time.Time
-	writeDeadline time.Time
+	readDeadline        time.Time
+	writeDeadline       time.Time
 
-	caPool *x509.CertPool
+	caPool              *x509.CertPool
 
 	onlyDHE             bool
 	onlyExports         bool
@@ -75,16 +75,16 @@ type Conn struct {
 	noSNI               bool
 	extendedRandom      bool
 
-	domain string
+	domain              string
 
 	// Encoding type
-	ReadEncoding string
+	ReadEncoding        string
 
 	// SSH
-	sshScan *SSHScanConfig
+	sshScan             *SSHScanConfig
 
 	// Errored component
-	erroredComponent string
+	erroredComponent    string
 }
 
 func (c *Conn) getUnderlyingConn() net.Conn {
@@ -232,6 +232,7 @@ func (c *Conn) makeHTTPRequest(config *HTTPConfig) (req *http.Request, encReq *H
 	} else {
 		userAgent = "Mozilla/5.0 zgrab/0.x"
 	}
+
 	req.Header.Set("User-Agent", userAgent)
 	encReq = new(HTTPRequest)
 	encReq.Endpoint = config.Endpoint
@@ -252,16 +253,16 @@ func (c *Conn) sendHTTPRequestReadHTTPResponse(req *http.Request, config *HTTPCo
 	var res *http.Response
 	if res, err = http.ReadResponse(reader, req); err != nil {
 		msg := err.Error()
-		if len(msg) > 1024*config.MaxSize {
-			err = errors.New(msg[0 : 1024*config.MaxSize])
+		if len(msg) > 1024 * config.MaxSize {
+			err = errors.New(msg[0 : 1024 * config.MaxSize])
 		}
 		return
 	}
 	var body []byte
 	if body, err = ioutil.ReadAll(res.Body); err != nil {
 		msg := err.Error()
-		if len(msg) > 1024*config.MaxSize {
-			err = errors.New(msg[0 : 1024*config.MaxSize])
+		if len(msg) > 1024 * config.MaxSize {
+			err = errors.New(msg[0 : 1024 * config.MaxSize])
 		}
 		return
 	}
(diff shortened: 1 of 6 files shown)


zlib/conn.go
```diff
@@ -41,20 +41,20 @@ const (
 // Implements the net.Conn interface
 type Conn struct {
 	// Underlying network connection
-	conn    net.Conn
-	tlsConn *ztls.Conn
-	isTls   bool
+	conn                net.Conn
```
zzma: Some auto-formatting from my IDE (IntelliJ w/ Go plugin). Let me know if we are using different auto-formatting standards.

zlib/conn.go
```diff
@@ -295,32 +298,83 @@ func (c *Conn) doProxy(config *HTTPConfig) error {
...
+			}
+
+			if httpRequest, httpResponse, err = c.makeAndSendHTTPRequest(config); err != nil {
+				return err
+			}
+
+			c.grabData.HTTP.RedirectRequests = append(c.grabData.HTTP.RedirectRequests, httpRequest)
+			c.grabData.HTTP.RedirectResponses = append(c.grabData.HTTP.RedirectResponses, httpResponse)
+
+		case http.StatusUseProxy:
```
zzma: This is a proxy redirect. Planning to implement this next. Right now it no-ops

zlib/grabber.go
```diff
@@ -238,6 +238,8 @@ func makeGrabber(config *Config) func(*Conn) error {
 			config.ErrorLog.Errorf("Conversation error with remote host %s: %s",
 				c.RemoteAddr().String(), err.Error())
 		}
+
+		c.Close()
```
zzma: From what I could tell, we were never explicitly closing any of our HTTP connections. This seems important because we are doing HTTP 1.1 in all our initial requests, which defaults to keep-alive connections.

zakird: We should also figure out how to make this work with only IP as long as the
redirect IP matches the original one pasted in (no domain was ever
specified)

zlib/conn.go
dadrian: I have my editor (Atom) configured to run gofmt with default parameters on save

zzma: that sounds like exactly what i should be using instead

zlib/http.go
```diff
@@ -130,3 +136,42 @@ func init() {
...
+
+	// Either explicit keep-alive or HTTP 1.1, which uses persistent connections by default
+	var keepAlive bool
+	if response.Headers["Connection"] != nil {
+		keepAlive = strings.EqualFold(response.Headers["Connection"].(string), "keep-alive")
+	} else {
+		keepAlive = response.VersionMajor == 1 && response.VersionMinor == 1
+	}
+
+	matchesHost := (targetUrl.Host == redirectUrl.Host) || (redirectUrl.Host == remoteIpAddress)
```
zzma: Added a conditional here to match host on both domain host and ip address

zzma:
Bump - let me know if you guys have any other comments

zlib/grabber.go
zzma: :+1: just added

zakird:
SO what about relative link. What happens if you redirect me to /about ?

zzma:
Hmm - this initially gave me the impression that the redirect location had to be an absolute URI: http://tools.ietf.org/html/rfc2616#section-14.30.

But looking into it a bit more, it seems like RFC2616 was replaced, and it could be a relative: https://tools.ietf.org/html/rfc7231#section-7.1.2

I'll go ahead and add relative URL handling

zlib/http.go
```diff
@@ -130,3 +136,44 @@ func init() {
...
+	if redirectUrlError != nil {
+		return false
+	}
+
+	remoteIpAddress, _, remoteIpAddressErr := net.SplitHostPort(conn.RemoteAddr().String())
+	if remoteIpAddressErr != nil {
+		return false
+	}
+
+	relativePath := redirectUrl.Host == ""
```
zzma: added check for relative path

zlib/conn.go
```diff
@@ -295,32 +298,82 @@ func (c *Conn) doProxy(config *HTTPConfig) error {
...
+		if location = httpResponse.Headers["location"].(string); location == "" {
+			return fmt.Errorf("No location found for %d response from %s (%s)", httpResponse.StatusCode, c.domain, c.RemoteAddr())
+		}
+
+		switch httpResponse.StatusCode {
+		case http.StatusMultipleChoices, http.StatusMovedPermanently, http.StatusFound, http.StatusSeeOther, http.StatusTemporaryRedirect:
+			if locationUrl, err := url.Parse(location); err != nil {
+				return err
+			} else {
+				config.Endpoint = locationUrl.RequestURI()
```
zzma: this will set the Endpoint correctly, regardless of absolute or relative URI

zzma:
https://upload.wikimedia.org/wikipedia/commons/thumb/1/17/Belize_Speed_Bump_Sign.JPG/1280px-Belize_Speed_Bump_Sign.JPG

zzma:
image

dadrian:
This looks pretty good to me based on reading the code, I'm going to play around with it some later today.

I'll try to get you access to a server to test from today.

zzma:
Thanks. I'm about 95% sure that some flavor of redirect may break it, since I haven't done much testing besides different redirect settings on a local server..

zakird:
If you want to test right away, you could test with the top 1000 Alex sites or some IPs that gave redirects in Censys. (look for status code 302).

zakird:
See https://censys.io/ipv4?q=80.http.get.status_code%3A302

zzma:
So I tested this with the top 100 and things look pretty good - there are a bunch of redirects resulting in 200s.  

Many of the redirects were simply redirecting from the main domain http request-uri to a sub-domain http request-uri (mostly www). For example, an HTTP request to request-uri "myDomain.com" would redirect to "www.myDomain.com". To get this working for most cases, I modified the redirect matching-domain check to see if the connection domain is a substring of the redirect location domain (https://github.com/zmap/zgrab/commit/5613e3a0eefeab6a7af9c103bbdf5b204109db07), assuming that myDomain.com and www.myDomain.com have the same IP address. However, there are 2 sites in the top 100 that have a different IP address for their bare domain i.e. myDomain.com and www.myDomain.com, so the redirect from myDomain.com to www.myDomain.com does not work when attempting it on the same connection. These fail quietly and just get the same redirect response a few times until max-redirects is reached. I think this is relatively uncommon and doesn't cause any real issues so it should be okay for now. When I implement multiple-connection redirects, I will make sure to come back and handle this more elegantly.

Next up, there are a few cases that we haven't covered that will be the focus of multiple-connection redirects. 1) HTTP->HTTPS redirects and 2) redirects to different domains. After that, if it's useful, I may look into proxy redirects.

zakird:
Sounds good! Let's run this against the full top milllion and a 1% scan and
see if we run into any issues. If not, and if @dadrian thinks that code
looks good, we'll get this merged in.

zzma:
Will do.

dadrian:
Code looks pretty good to me---at least as good as the old HTTP code and probably better :).

Assuming it holds up for full Alexa and 1% scans, I'm happy to merge. Ping me if you have any issues with the 1% scans. We'll also want to double check that we didn't break anything in our regular TLS scans (unlikely, but worth checking).

Out of curiosity, do we currently handle HTTPS -> HTTPS redirects if we pass the --tls and --http-max-redirects together?

zzma:
The https-->https redirects work. 

Testing with 1% scan reveals a deadlock situation with the default 1000 workers, which I'm looking into now

dadrian:
@zzma nice!

zzma:
Turns out it wasn't a deadlock situation - just a interface conversion error that was hidden in a bunch of noise. Tested it out again on top 1million and 1% - the only issue seems to be that the output file is pretty huge. For top 1 million, the output came out to 43Gb. For 1% scan, I had to redirect output to /dev/null. Is this a problem? If so, maybe we can squelch the response body (which contributes to a significant amt of data) and just display header information?

zakird:
We used to have a max amount of content. Does that still work?

zzma:
yes good call - the max size still works. the 43Gb top 1 million scan was using the default 256Kb response size. Looks like we can just tweak that --http-max-size parameter if the output size gets out of hand.

zzma:
@dadrian - is this good to go? 

Also FWIW - 35% of the top 1 million had a redirect response, and we were surprisingly able to handle about 66% of those despite our initial strict redirect requirements.

dadrian:
Merged by 3a438bfe7c9605e59edb7e873a79f2d7a4a7444d

PR #71, opened 2015-11-04
Preliminary support for ISCSI Scanning

benthor:
This still needs better integration with the rest of the codebase (e.g., config handling), some additional logic to prevent conflicting flags and a more configurable handshake handling. This code was hacked together in just a few hours and is not very clean and lacks a lot of features.

### DIFF
--- a/zlib/conn.go
+++ b/zlib/conn.go
@@ -32,8 +32,10 @@ import (
 	"github.com/zmap/zgrab/ztools/ssh"
 	"github.com/zmap/zgrab/ztools/util"
 	"github.com/zmap/zgrab/ztools/x509"
+	"github.com/zmap/zgrab/ztools/zlog"
 	"github.com/zmap/zgrab/ztools/ztls"
 	"github.com/zmap/zgrab/ztools/iscsi"
+	"io"
 )
 
 var smtpEndRegex = regexp.MustCompile(`(?:^\d\d\d\s.*\r\n$)|(?:^\d\d\d-[\s\S]*\r\n\d\d\d\s.*\r\n$)`)
@@ -208,8 +210,8 @@ func (c *Conn) HTTP(config *HTTPConfig) error {
 	return c.doHTTP(config)
 }
 
-func (c *Conn) makeHTTPRequest(config *HTTPConfig) (req *http.Request, encReq *HTTPRequest, err error) {
-	if req, err = http.NewRequest(config.Method, "", nil); err != nil {
+func (c *Conn) makeHTTPRequest(endpoint string, httpMethod string, userAgent string) (req *http.Request, encReq *HTTPRequest, err error) {
+	if req, err = http.NewRequest(httpMethod, "", nil); err != nil {
 		return
 	}
 	url := new(url.URL)
@@ -221,27 +223,30 @@ func (c *Conn) makeHTTPRequest(config *HTTPConfig) (req *http.Request, encReq *H
 	}
 	url.Host = host
 	req.Host = host
-	req.Method = config.Method
+	req.Method = httpMethod
 	req.Proto = "HTTP/1.1"
 	if c.isTls {
 		url.Scheme = "https"
 	} else {
 		url.Scheme = "http"
 	}
-	url.Path = config.Endpoint
+	url.Path = endpoint
 	req.URL = url
-	var userAgent string
-	if len(config.UserAgent) > 0 {
-		userAgent = config.UserAgent
-	} else {
+
+	if len(userAgent) <= 0 {
 		userAgent = "Mozilla/5.0 zgrab/0.x"
 	}
+
 	req.Header.Set("User-Agent", userAgent)
 	encReq = new(HTTPRequest)
-	encReq.Endpoint = config.Endpoint
-	encReq.Method = config.Method
+	encReq.Endpoint = endpoint
+	encReq.Method = httpMethod
 	encReq.UserAgent = userAgent
 	return req, encReq, nil
+}
+
+func (c *Conn) makeHTTPRequestFromConfig(config *HTTPConfig) (req *http.Request, encReq *HTTPRequest, err error) {
+	return c.makeHTTPRequest(config.Endpoint, config.Method, config.UserAgent)
 }
 
 func (c *Conn) sendHTTPRequestReadHTTPResponse(req *http.Request, config *HTTPConfig) (encRes *HTTPResponse, err error) {
@@ -272,6 +277,8 @@ func (c *Conn) sendHTTPRequestReadHTTPResponse(req *http.Request, config *HTTPCo
 	encRes = new(HTTPResponse)
 	encRes.StatusCode = res.StatusCode
 	encRes.StatusLine = res.Proto + " " + res.Status
+	encRes.VersionMajor = res.ProtoMajor
+	encRes.VersionMinor = res.ProtoMinor
 	encRes.Headers = HeadersFromGolangHeaders(res.Header)
 	var bodyOutput []byte
 	if len(body) > 1024*config.MaxSize {
@@ -289,7 +296,7 @@ func (c *Conn) sendHTTPRequestReadHTTPResponse(req *http.Request, config *HTTPCo
 }
 
 func (c *Conn) doProxy(config *HTTPConfig) error {
-	req, encReq, err := c.makeHTTPRequest(config)
+	req, encReq, err := c.makeHTTPRequestFromConfig(config)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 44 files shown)


ztools/iscsi/scan.go
```diff
@@ -0,0 +1,117 @@
...
+		"HeaderDigest=None",
+		"DataDigest=None",
+		"DefaultTime2Wait=2",
+		"DefaultTime2Retain=0",
+		"IFMarker=No",
+		"OFMarker=No",
+		"ErrorRecoveryLevel=0",
+		"MaxRecvDataSegmentLength=32768",
+	)
+	header := fmt.Sprintf("438700000000%04x00023D00000000000000000000000000000000010000000000000000000000000000000000000000", len(body)/2)
```
dadrian: Where is the data in this header coming from?

ztools/iscsi/scan.go
```diff
@@ -0,0 +1,117 @@
...
+		authlog.HadError = true
+		return authlog, err
+	}
+	var haderror error
+	for _, bs := range iscsiTargetRegex.FindAllSubmatch(res, -1) {
+		//for i, b := range bs {
+		//fmt.Println(i, string(b), len(bs), n)
+	//}
+
+		conn2, err := net.Dial("tcp", conn.RemoteAddr().String())
```
dadrian: Does this have to be a new connection?

ztools/iscsi/scan.go
```diff
@@ -0,0 +1,117 @@
...
+	conn.Write(data)
+	res = make([]byte, 1024)
+	_, err = conn.Read(res) //c.readUntilRegex(res, iscsiTargetNameRegex)
+	//c.grabData.Banner = string(res[0:n])
+	if err != nil {
+		authlog.HadError = true
+		return authlog, err
+	}
+	var haderror error
+	for _, bs := range iscsiTargetRegex.FindAllSubmatch(res, -1) {
```
dadrian: If we're making more connections, we need to set an upper bound on how many of these we process, so we don't DoS ourselves or the remote host.

ztools/iscsi/scan.go
```diff
@@ -0,0 +1,117 @@
...
+			authlog.Targets = append(authlog.Targets, Target{string(bs[1]), string(bs[2]), true, false})
+		} else {
+			authlog.Targets = append(authlog.Targets, Target{string(bs[1]), string(bs[2]), false, false})
+		}
+		
+	}
+	if haderror != nil {
+		authlog.HadError = true
+	}
+	return authlog, haderror
```
dadrian: What's the point of this `HadError` bool? Why not just pass an error object around?

ztools/iscsi/scan.go
```diff
@@ -0,0 +1,117 @@
...
+			"FirstBurstLength=262144", //15
+			"MaxOutstandingR2T=1", //16
+			"MaxConnections=1", //17
+			"DataPDUInOrder=Yes", //18
+			"DataSequenceInOrder=Yes", //19
+			"MaxRecvDataSegmentLength=262144", //20
+		)
+		header = fmt.Sprintf("438700000000%04x00023D00000000000000000000000000000000010000000000000000000000000000000000000000", len(body)/2)
+		data = []byte{}
+		fmt.Sscanf(header+body+"0000", "%x", &data)
```
dadrian: I'd rather see this done with `Marshal()` and `Unmarshal()` functions on structs (see the SSH or TLS libraries as an example)

ztools/iscsi/log.go
```diff
@@ -0,0 +1,14 @@
+package iscsi
+
+type Target struct {
+	Name string `json:"target_name"`
```
dadrian: Is there a reason these strings aren't omitempty? Does empty string have any meaning here?

dadrian:
I'm a little confused about what exactly this module is doing.

1) Why do we need more than one connection per host?
2) Is this module attempting to login? What are AuthTargets?
3) Do you have an example output and a host we can test against?

Beyond that, as far as integration into the rest of ZGrab goes, the command line flags, configuration, etc. look fine. All my concerns have to do with the scanner itself.

ztools/iscsi/scan.go
benthor: From a proof-of-concept script which I ported, which in turn mostly reflected stuff observed in wireshark. I agree that much of this should probably be made configurable.

ztools/iscsi/scan.go
benthor: It does. I couldn't make it work with a single connection (and the proof of concept script which I ported did it the same way)

ztools/iscsi/scan.go
benthor: We are only making more connections on targets which the other host reports. But I agree, some upper bound would make sense

ztools/iscsi/scan.go
benthor: Because authlog is to be marshalled into json where we want to have this information

ztools/iscsi/scan.go
benthor: I agree, but (as it must be apparent), I had to cut a lot of corners to make this work. (My colleague asked me to get as much as possible working until the end of the day, as he urgently needs to collect data)

ztools/iscsi/log.go
benthor: These strings can be made omitempty. My bad

benthor:
Re 1) I couldn't make it work with just one connection. My iscsi background is minimal and I was on a tight time budget, porting a python script to this code
Re 2) It does not attempt login. It merely scans the server reply for enabled or disabled authentication before closing the connection.
Re 3) Yes, see the email where I originally submitted the patch to you

zakird:
To be clear, while we're happy to help and I understand you're on a deadline, we cannot be sending out binary strings that we don't know the meaning of nor cutting corners like this. This code will run against every host on the IPv4 address space and we need to be able to guarantee that the code is safe. Let's work on cleaning it up, testing it against a variety of services, and then we can work on getting it merged.

benthor:
The code is a 1:1 emulation of an iSCSI session as observed in wireshark, so it should be safe. That said, I'm currently working my way along the RFC and creating an actual (partial) bottom-up protocol implementation for the stuff we are interested in. I hope to have something ready later today.

benthor:
I largely rewrote the module. Please have a look

dadrian:
I'll try to get to this over the weekend.

dadrian:
Code looks pretty good so far, but I'm still worried about multiple connections per host. @benthor which RFC is this based on?

benthor:
https://tools.ietf.org/html/rfc7143

benthor:
I have just had a detailed look at the iSCSI state machine and it is a mess. I had hoped that it would be possible to at least scan _one_ target with the established connection but that does not appear to be the case. 

> Discovery session - a session only opened for target discovery. The target MUST ONLY accept Text Requests with the SendTargets key and a Logout Request with reason "close the session".  All other requests MUST be rejected.

So to determine whether there are targets on the host which don't require authentication, we can't use the discovery session. We have to establish a fully-featured session to see the relevant headers. Here, the iSCSI state machine causes problems. The only point in which we can see these headers is in the IN_LOGIN state. We can not recycle that state however. Firstly, we have bound our session to that target and secondly, any error state leads to a state _before_ session establishment, forcing us to re-establish another TCP connection. (See RFC7143, section 8).

(I even had a look at how the nmap iscsi plugin does it. They too open a connection for each target they discovered.)

zlib/conn.go
```diff
@@ -618,3 +622,12 @@ func (c *Conn) SSHHandshake() error {
 	c.grabData.SSH = handshakeLog
 	return err
 }
+
+func (c *Conn) ISCSIScan(config *iscsi.ISCSIConfig) error {
+	data, err := iscsi.Scan(c.getUnderlyingConn(), config)
+	if err != nil {
```
dadrian: Are you sure this doesn't throw away data? What if we timeout halfway through, we still want to record whatever we were able to get.

ztools/iscsi/iscsi.go
```diff
@@ -0,0 +1,332 @@
...
+	TotalAHSLength    byte
+	DataSegmentLength [3]byte
+}
+
+type TextParameter struct {
+	Key   string
+	Value string
+}
+
+func (t *TextParameter) GetLength() {
```
dadrian: Why is this empty?

ztools/iscsi/iscsi.go
```diff
@@ -0,0 +1,332 @@
...
+	// empty for now
+}
+
+type BasicHeader struct {
+	Opcode            byte
+	Flags             byte
+	VersionMax        byte
+	VersionMin        byte
+	TotalAHSLength    byte
+	DataSegmentLength [3]byte
```
dadrian: What's the rational for using a [3]byte vs a []byte vs a bounded uint32? Either way you have validation that has to be done.

ztools/iscsi/iscsi.go
```diff
@@ -0,0 +1,332 @@
...
+func (p PDU) MarshalBinary() ([]byte, error) {
+	buf := new(bytes.Buffer)
+	//copy(p.Header.GetLength(), length)
+	//p.Header.GetLength()
+	err := binary.Write(buf, binary.BigEndian, p.Header)
+	if err != nil {
+		return buf.Bytes(), err
+	}
+	for _, param := range p.Data.Data {
+		_, err = buf.Write([]byte(param.Key + "=" + param.Value + "\u0000"))
```
dadrian: What does golang interpret the `"\u0000"` as? Why not just operate on bytes directly here?

ztools/iscsi/iscsi.go
```diff
@@ -0,0 +1,332 @@
...
+		if err != nil {
+			return buf.Bytes(), err
+		}
+	}
+
+	for i := 0; i < p.Data.L%4; i++ {
+		buf.Write([]byte{0})
```
dadrian: I assume this is something out of the RFC? Padding?

zlib/conn.go
benthor: Good catch.

ztools/iscsi/iscsi.go
benthor: Header had to be an "easy to implement" interface which allowed me to keep pdu.Marshal and pdu.Unmarshal in a generic way but supports all the different header types. Each header type "implements" this interface since they are only made up of bytes or byte arrays which binary.Read and binary.Write know how to (de)serialize. It is a bit weird to do it this way but it works just fine.

ztools/iscsi/iscsi.go
benthor: iSCSI uses exactly 3 bytes to denote the length of the data section of each PDU. binary.Read and binary.Write unfortunately can not handle slices and uint32 would have meant 4 bytes on the wire not 3, which would have broken things.

ztools/iscsi/iscsi.go
benthor: \u0000 is the "zero" unicode code point, the resulting byte is 0. I could of course have written this as `_, err = buf.Write(append([]byte(param.Key + "=" + param.Value), 0))` but I felt this was more readable

ztools/iscsi/iscsi.go
benthor: Exactly. Padding to n % 4 == 0 bytes.

dadrian:
@benthor can you fix the merge conflicts? I'd like to test this locally.

Also, can you explain what you would like to learn from these scans? We don't do scans where we attempt to log into services using default passwords. How would you differentiate the NoAuth login requests from logging in with a default password?

Beyond that, do you know what these login attempts will look like in server-side logs or IDS systems?

benthor:
This is not attempting any default logins. The way that iSCSI is designed is that any authentication happens _after_ the session is established. We terminate the connection as soon as we learn whether the host has authentication enabled for a target or not. The host tells us this in their response when we _initiate_ login to any target which was discovered. We never actually follow through with the login (even if the host tells us that for a given target, no authentication is required). There is simply no code for anything beyond reading the first reply. 

The server side logs will reflect this fact. They will not even show any "failed" login attempts, since we terminate our sessions before we even enter the authentication state.

To equate this with an HTTP-scanning example: We are doing the iSCSI equivalent of GET requests where we distinguish 200 (ok) and 301 (e.g., redirect to https) responses. We will never even trigger an error 403, because we are _not following up on_ any 301 responses.

Another example: Talking to an SMTP server: We close the connection as soon as we read "STARTTLS" and make a tick mark on our clipboard that the server supposedly enforces crypto.

dadrian:
I enabled iSCSI on a test host with the following config straight from the BSD example configuration (https://www.freebsd.org/doc/handbook/network-iscsi.html).

This should be discoverable, but the only output I get is here: https://gist.github.com/dadrian/8087fa01e43a7c421e49. Shouldn't the targets have been enumerated? Also, why are you including `ip` as a subfield of `iscsi`?

benthor:
What do you get when you run: `sudo iscsiadm -m discovery -t st -p 192.168.59.103`? 

Also, have you tried it with our test-host (ip in the email)? You are free to use that as well. This should definitely have output.

The second "ip" field is somewhat redundant, I agree. When I wrote the logger, I wasn't aware that "ip" was already included in the output.

benthor:
I fixed my marshaler to get rid of the redundant "ip" field

dadrian:
The output from `iscsiadm` is

```
192.168.59.103:3260,-1 iqn.2012-06.com.example:target0
```

dadrian:
When I run it on your example host, the zgrab output looks better and I get the following output:
```
xxx.xxx.xxx.xxx:3260,1 iqn.2001-04.com.example:storage.lun1
xxx.xxx.xxx.xxx:3260,1 iqn.2001-04.com.example:storage.lun0
```

benthor:
I am currently in the process of reproducing and debugging this error using my own freebsd vm. I'll report back once I know more.

benthor:
Please test again.

dadrian:
OK, so the target appears now.

```
{
  "ip": "192.168.59.103",
  "timestamp": "2015-11-15T14:33:48-05:00",
  "data": {
    "iscsi": {
      "targets": [
        {
          "target_name": "iqn.2012-06.com.example:target0",
          "target_address": "N/A",
          "authentication_disabled": true,
          "had_error": false
        }
      ],
      "had_error": false
    }
  }
}
```
However, I'm little confused about what the purpose of the target address field is, and I want to make sure we're conveying the correct information in the best possible manner. First off, rather than set target address to N/A, we probably just want to set the field to `omitempty`, and then not set a value for that string.

Second, when I run the scanner against your host, I get the following output:
```
{
  "ip": "xxx.xxx.xxx.xxx",
  "timestamp": "2015-11-15T14:32:40-05:00",
  "data": {
    "iscsi": {
      "targets": [
        {
          "target_name": "iqn.2001-04.com.example:storage.lun1",
          "target_address": "xxx.xxx.xxx.xxx:3260,1",
          "authentication_disabled": false,
          "had_error": false
        },
        {
          "target_name": "iqn.2001-04.com.example:storage.lun0",
          "target_address": "xxx.xxx.xxx.xxx:3260,1",
          "authentication_disabled": true,
          "had_error": false
        }
      ],
      "had_error": false
    }
  }
}
```
I may have just missed this earlier, but I don't previously remember the `,1` at the end. What's the significance of the `,1`? I see that `icsisadm` also outputs either a 1 or a -1 at the end of each target (see my earlier pastes). At any rate, it looks like `target_address` should probably be it's own struct that looks something like
```
type TargetAddress struct {
    IP string
    Port uint16
    Flag int
}
```
Thoughts?

benthor:
That is a good point. I'll look into it either later today or tomorrow

dadrian:
@benthor any updates?

benthor:
Sorry I haven't yet gotten around to merge my local changes. (Was sick for a week, had to catch up on other work and currently travelling...)

I hope to get back to this once I have a free moment.

dadrian:
Due to lack of activity, I'm closing this pull request. Feel free to reopen in the future if you'd like to work towards getting this merged.

PR #86, opened 2015-11-13
use terminal response as top level response

zzma:
@zakird @dadrian 

Minor tweak to HTTP redirects. Previously we were storing the initial HTTP response (aka a 300 response) as the top-level HTTPRequestResponse.response, and the terminal 200 response was being stored as the last element of HTTPRequestResponse.redirect_responses. I've change it so that HTTPRequestResponse.response now contains the terminal response at the end of max-redirects, and moved the initial http response to be the first item of HTTPRequestResponse.redirect_responses.

One concern I have is that the requests and responses are misaligned. HTTPRequestResponse.request does not directly return HTTPRequestResponse.response. Also the HTTPRequestResponse.redirect_requests and HTTPRequestResponse.redirect_responses are off-by-one. Are we okay with this?

### DIFF
--- a/zgrab_schema.py
+++ b/zgrab_schema.py
@@ -330,6 +330,8 @@ zgrab_http_request = SubRecord({
 })
 
 zgrab_http_response = SubRecord({
+    "version_major":Integer(),
+    "version_minor":Integer(),
     "status_code":Integer(),
     "status_line":AnalyzedString(),
     "body":HTML(),
--- a/zgrab_schema.py
+++ b/zgrab_schema.py
@@ -337,11 +337,16 @@ zgrab_http_response = SubRecord({
     "headers":zgrab_http_headers
 })
 
+zgrab_http_request_response = SubRecord({
+    "request":zgrab_http_request,
+    "response":zgrab_http_response
+})
+
 zgrab_http = Record({
     "data":SubRecord({
       "http":SubRecord({
-        "request":zgrab_http_request,
-        "response":zgrab_http_response
+        "response":zgrab_http_response,
+        "request_response_chain":ListOf(zgrab_http_request_response)
       })
     })
 }, extends=zgrab_base)
--- a/zlib/http.go
+++ b/zlib/http.go
@@ -76,13 +76,16 @@ type HTTPResponse struct {
 	BodySHA256   []byte      `json:"body_sha256,omitempty"`
 }
 
+type HTTP struct {
+	ProxyRequest         *HTTPRequest           `json:"connect_request,omitempty"`
+	ProxyResponse        *HTTPResponse          `json:"connect_response,omitempty"`
+	Response             *HTTPResponse          `json:"response,omitempty"`
+	RequestResponseChain []*HTTPRequestResponse `json:"request_response_chain,omitempty"`
+}
+
 type HTTPRequestResponse struct {
-	ProxyRequest      *HTTPRequest    `json:"connect_request,omitempty"`
-	ProxyResponse     *HTTPResponse   `json:"connect_response,omitempty"`
-	Request           *HTTPRequest    `json:"request,omitempty"`
-	Response          *HTTPResponse   `json:"response,omitempty"`
-	RedirectRequests  []*HTTPRequest  `json:"redirect_requests,omitempty"`
-	RedirectResponses []*HTTPResponse `json:"redirect_responses,omitempty"`
+	Request  *HTTPRequest  `json:"request,omitempty"`
+	Response *HTTPResponse `json:"response,omitempty"`
 }
 
 func init() {
--- a/zlib/conn.go
+++ b/zlib/conn.go
@@ -297,7 +297,7 @@ func (c *Conn) doProxy(config *HTTPConfig) error {
 		return err
 	}
 	if c.grabData.HTTP == nil {
-		c.grabData.HTTP = new(HTTPRequestResponse)
+		c.grabData.HTTP = new(HTTP)
 	}
 	c.grabData.HTTP.ProxyRequest = encReq
 	req.Method = "CONNECT"
@@ -317,17 +317,19 @@ func (c *Conn) doProxy(config *HTTPConfig) error {
 
 func (c *Conn) doHTTP(config *HTTPConfig) error {
 	if c.grabData.HTTP == nil {
-		c.grabData.HTTP = new(HTTPRequestResponse)
+		c.grabData.HTTP = new(HTTP)
 	}
 
 	var httpResponse, terminalResponse, previousResponse *HTTPResponse
 	var httpRequest *HTTPRequest
+	var requestResponse *HTTPRequestResponse
 	var err error
 	if httpRequest, httpResponse, err = c.makeAndSendHTTPRequest(config); err != nil {
 		return err
 	}
 
-	c.grabData.HTTP.Request = httpRequest
+	requestResponse = new(HTTPRequestResponse)
+	requestResponse.Request = httpRequest
 	terminalResponse = httpResponse
 
 	for redirectCount := 0; httpResponse.isRedirect() && httpResponse.canRedirectWithConn(c) && redirectCount < config.MaxRedirects; redirectCount++ {
@@ -362,8 +364,10 @@ func (c *Conn) doHTTP(config *HTTPConfig) error {
 				return err
 			}
 
-			c.grabData.HTTP.RedirectRequests = append(c.grabData.HTTP.RedirectRequests, httpRequest)
-			c.grabData.HTTP.RedirectResponses = append(c.grabData.HTTP.RedirectResponses, previousResponse)
+			requestResponse.Response = previousResponse
+			c.grabData.HTTP.RequestResponseChain = append(c.grabData.HTTP.RequestResponseChain, requestResponse)
+			requestResponse = new(HTTPRequestResponse)
+			requestResponse.Request = httpRequest
 
 			terminalResponse = httpResponse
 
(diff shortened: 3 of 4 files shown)


zakird:
Is there a reason we can't have one array that contains both? matching up
seems pretty obnoxious.

zzma:
Do you mean a single array called something like redirectRequestResponseChain that contains [response1, request2, response2, request3, response3 (terminal)]? It seems like the response data is of most practical interest, and having to parse through the array each time and filter out requests would be annoying. That's why I wanted to store them separately. Do you agree?

To handle the issue of matching up, we could potentially duplicate data (i.e. the initial request / terminal response) so that the arrays match up nicely, but I'm certain that duplicating the terminal 200 responses in particular will significantly increase the size of output data. Alternatively, we could match up by filling in null for the data we don't want to duplicate and make it clear in the documentation.

dadrian:
Could we do something along the lines of [ { request, response }, {request, response } ], plus the top-level information?

zzma:
@dadrian - we definitely could. Would we still want the HTTPRequestResponse.Response convenience field in this case? Or make users find that by going to the end of the request,response array? How much existing infrastructure / people out there do you think are relying on the HTTPRequestResponse.Request / HTTPRequestResponse.Response fields?

zakird:
What are in each of those? I think the terminal one should be its own,
separate from the chain to get there.

zzma:
Right now on master we are storing the the initial request in HTTPRequestResponse.Request and the initial response in HTTPRequestResponse.Response. This PR places the terminal response in HTTPRequestResponse.Response. 

In addressing how to store the other requests/responses, I am making the assumption that we want to avoid duplication, especially duplication of the terminal response because the 200 responses are usually an order of magnitude larger than the rest of the data combined. It seems like one solution for redirects is to add a new array called requestResponseChain that stores {request, response} pairs, but leave out the initial request and terminal response and leave those parts null in the corresponding {request, response} pair. Thoughts?

zakird:
This sounds fine. I'm not sure why we'd have initial response be outside of
the array though. I'd only keep terminal separate.

zzma:
Bump. I renamed `HTTPRequestResponse` which was storing everything HTTP related to `HTTP`. Also removed `Request`, `RedirectResponses` and `RedirectRequests` in favor of `RedirectResponseChain`. `RedirectResponseChain` is an array of the newly created `HTTPRequestResponse` objects, which now contain only a `HTTPRequest` / `HTTPResponse` pair

zlib/http.go
```diff
@@ -76,13 +76,16 @@ type HTTPResponse struct {
...
+	ProxyResponse        *HTTPResponse          `json:"connect_response,omitempty"`
+	Response             *HTTPResponse          `json:"response,omitempty"`
+	RequestResponseChain []*HTTPRequestResponse `json:"request_response_chain,omitempty"`
+}
+
 type HTTPRequestResponse struct {
-	ProxyRequest      *HTTPRequest    `json:"connect_request,omitempty"`
-	ProxyResponse     *HTTPResponse   `json:"connect_response,omitempty"`
-	Request           *HTTPRequest    `json:"request,omitempty"`
-	Response          *HTTPResponse   `json:"response,omitempty"`
```
dadrian: Did we previously ever output this field? It looks like we might have lost our original request in the new format?

dadrian:
We're going to have to update schema.py with the new fields. From what I can tell, the type of any existing field did not change, we just changed semantics to account for having a redirect chain?
Overall, changes look good though.

dadrian:
Can you post an example of both the old output, and the new output?

zlib/http.go
zzma: Yes - we previously output both the initial request and the initial response (and proxy request/response, if present). The new format loses the explicit `Request` field, but the initial request can be found in the first HTTPRequestReponse element in the HTTP.RequestResponseChain. The somewhat confusing part is that I renamed the original HTTPRequestResponse struct to simply HTTP, and then I also created a new struct called HTTPRequestResponse that only contains a Request/Response pair.

zzma:
We changed the name of the HTTPRequestResponse struct to HTTP, removed the Request field from the new HTTP struct, and added a redirect chain to the new HTTP struct. I'll look into updating `zgrab_schema.py`. Out of curiosity - what is that file used for?

Old output: 

```
{"ip":"23.21.237.247","domain":"pbs.org","timestamp":"2015-11-14T18:13:21-05:00","data":{"http":{"request":{"method":"GET","endpoint":"/","user_agent":"Mozilla/5.0 zgrab/0.x"},"response":{"version_major":1,"version_minor":1,"status_code":301,"status_line":"HTTP/1.1 301 Moved Permanently","headers":{"content_length":"227","content_type":"text/html; charset=iso-8859-1","location":"http://www.pbs.org/","server":"Apache"},"body":"\u003c!DOCTYPE HTML PUBLIC \"-//IETF//DTD HTML 2.0//EN\"\u003e\n\u003chtml\u003e\u003chead\u003e\n\u003ctitle\u003e301 Moved Permanently\u003c/title\u003e\n\u003c/head\u003e\u003cbody\u003e\n\u003ch1\u003eMoved Permanently\u003c/h1\u003e\n\u003cp\u003eThe document has moved \u003ca href=\"http://www.pbs.org/\"\u003ehere\u003c/a\u003e.\u003c/p\u003e\n\u003c/body\u003e\u003c/html\u003e\n","body_sha256":"PhovM2HBtuxMulpzFmqz/J/awAWg+S+6MP4OqWKTlOk="},"redirect_requests":[{"method":"GET","endpoint":"/","user_agent":"Mozilla/5.0 zgrab/0.x"}],"redirect_responses":[{"version_major":1,"version_minor":1,"status_code":200,"status_line":"HTTP/1.1 200 OK","headers":{"accept_ranges":"bytes","content_length":"87","content_type":"text/html","server":"Apache","unknown":[{"key":"x_pbs_fwsrvname","value":"admin-pbs-org"},{"key":"x_pbs_fwsrvip","value":"10.168.214.125"}],"vary":"Accept-Encoding"},"body":"\u003c!--\nThis is just a placeholder.\nThe actual homepage is proxied to another server.\n--\u003e\n","body_sha256":"TQD5f6voLlSxhA4LZmGXGu2PD7eYr+v1ZeKn0pUL5wM="}]}}}
```

New output:

```
{"ip":"23.21.237.247","domain":"pbs.org","timestamp":"2015-11-14T18:25:02-05:00","data":{"http":{"response":{"version_major":1,"version_minor":1,"status_code":200,"status_line":"HTTP/1.1 200 OK","headers":{"accept_ranges":"bytes","content_length":"87","content_type":"text/html","server":"Apache","unknown":[{"key":"x_pbs_fwsrvname","value":"admin-pbs-org"},{"key":"x_pbs_fwsrvip","value":"10.168.214.125"}],"vary":"Accept-Encoding"},"body":"\u003c!--\nThis is just a placeholder.\nThe actual homepage is proxied to another server.\n--\u003e\n","body_sha256":"TQD5f6voLlSxhA4LZmGXGu2PD7eYr+v1ZeKn0pUL5wM="},"request_response_chain":[{"request":{"method":"GET","endpoint":"/","user_agent":"Mozilla/5.0 zgrab/0.x"},"response":{"version_major":1,"version_minor":1,"status_code":301,"status_line":"HTTP/1.1 301 Moved Permanently","headers":{"content_length":"227","content_type":"text/html; charset=iso-8859-1","location":"http://www.pbs.org/","server":"Apache"},"body":"\u003c!DOCTYPE HTML PUBLIC \"-//IETF//DTD HTML 2.0//EN\"\u003e\n\u003chtml\u003e\u003chead\u003e\n\u003ctitle\u003e301 Moved Permanently\u003c/title\u003e\n\u003c/head\u003e\u003cbody\u003e\n\u003ch1\u003eMoved Permanently\u003c/h1\u003e\n\u003cp\u003eThe document has moved \u003ca href=\"http://www.pbs.org/\"\u003ehere\u003c/a\u003e.\u003c/p\u003e\n\u003c/body\u003e\u003c/html\u003e\n","body_sha256":"PhovM2HBtuxMulpzFmqz/J/awAWg+S+6MP4OqWKTlOk="}},{"request":{"method":"GET","endpoint":"/","user_agent":"Mozilla/5.0 zgrab/0.x"}}]}}}
```

zgrab_schema.py
```diff
@@ -337,11 +337,16 @@
...
+    "response":zgrab_http_response
+})
+
 zgrab_http = Record({
     "data":SubRecord({
       "http":SubRecord({
-        "request":zgrab_http_request,
-        "response":zgrab_http_response
+        "response":zgrab_http_response,
+        "request_response_chain":ListOf(zgrab_http_request_response)
```
zzma: @dadrian - am I doing this right? Also - what's the best way for me to test this?

dadrian:
@zzma the names of the Go structs don't matter, it's the names of the fields in the JSON output we need to keep track of. We actually strongly-type all the JSON data, as it eventually gets propagated in to Censys.

zgrab_schema.py
dadrian: That looks good. The schema file is a DSL called zschema. You can install it with `pip install zschema` and then have it validate that a JSON file matches a given schema. Ask @zakird or @amirian28 for more details about how to use it, it's quite possibly the least documented tool that we have.

dadrian: Also defined in this file there's a `zgrab_http_proxy` object (or something similarly named). Can you make sure that it is still to date?

zzma: Yup - the `zgrab_http_proxy` is up-to-date

zakird:
can someone paste me an example record?

zakird:
This looks good to me.

dadrian:
Have we always had `version_major` and `version_minor`? I don't see it listed in zgrab_schema.py.

zzma:
We have not - I added them in as part of the HTTP redirect stuff because we needed to parse that information anyways in order to determine if the connection was Keep-Alive by default, which is true for HTTP 1.1

dadrian:
@zzma, ah ok. That's reasonable. Can you update the schema accordingly?

zzma:
Done!

PR #91, opened 2015-11-20
Zzma/telnet banner grab

zzma:
@zakird @dadrian 

Basic telnet banner grab. I saw an error once, but wasn't able to capture it, but I've tested about 5 times since on a 1% scan and have not been able to reproduce the error.

### DIFF
--- a/ztools/telnet/log.go
+++ b/ztools/telnet/log.go
@@ -1,0 +1,13 @@
+/*
+ * ZGrab Copyright 2015 Regents of the University of Michigan
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy
+ * of the License at http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+package telnet
+
+type TelnetLog struct {
+	Banner string `json:"banner,omitempty"`
+}
--- a/ztools/telnet/log.go
+++ b/ztools/telnet/log.go
@@ -15,5 +15,7 @@
 package telnet
 
 type TelnetLog struct {
-	Banner string `json:"banner,omitempty"`
+	Banner          string   `json:"banner,omitempty"`
+	SupportedOpts   []string `json:"supported_opts,omitempty"`   //DO and WILL option codes
+	UnsupportedOpts []string `json:"unsupported_opts,omitempty"` // DONT and WONT option codes
 }
--- a/ztools/telnet/log.go
+++ b/ztools/telnet/log.go
@@ -4,6 +4,12 @@
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not
  * use this file except in compliance with the License. You may obtain a copy
  * of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+ * implied. See the License for the specific language governing
+ * permissions and limitations under the License.
  */
 
 package telnet
--- a/ztools/telnet/telnet.go
+++ b/ztools/telnet/telnet.go
@@ -45,7 +45,7 @@ func GetTelnetBanner(logStruct *TelnetLog, conn net.Conn, maxReadSize int) (err 
 	numBytes := len(buffer)
 	rounds := int(math.Ceil(float64(maxReadSize) / READ_BUFFER_LENGTH))
 	count := 0
-	for numBytes != 0 && count < rounds {
+	for numBytes != 0 && count < rounds && numBytes == READ_BUFFER_LENGTH {
 
 		numBytes, err = conn.Read(buffer)
 		if err != nil && err != io.EOF {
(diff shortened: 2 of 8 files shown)


zakird:
Do we lose clients if we don't have anything we that we support? Also, do
we want to log what servers support?

zakird:
Err, servers

ztools/telnet/log.go
```diff
@@ -0,0 +1,13 @@
+/*
+ * ZGrab Copyright 2015 Regents of the University of Michigan
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy
+ * of the License at http://www.apache.org/licenses/LICENSE-2.0
+ */
```
dadrian: Can you add the warranty clause here? It should be in the other files, I went through and added it to all files a couple weeks ago.

ztools/telnet/telnet.go
```diff
@@ -0,0 +1,94 @@
+/*
+ * ZGrab Copyright 2015 Regents of the University of Michigan
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy
+ * of the License at http://www.apache.org/licenses/LICENSE-2.0
+ */
+
```
dadrian: Warranty clause.

ztools/telnet/telnet.go
```diff
@@ -0,0 +1,94 @@
...
+package telnet
+
+import (
+	"bytes"
+	"errors"
+	"net"
+)
+
+var (
```
dadrian: This should be a const block.

ztools/telnet/telnet.go
```diff
@@ -0,0 +1,94 @@
...
+	DONT               = byte(254)
+	DO                 = byte(253)
+	WONT               = byte(252)
+	WILL               = byte(251)
+	IAC_CMD_LENGTH     = 3 // IAC commands take 3 bytes (inclusive)
+	READ_BUFFER_LENGTH = 8192
+)
+
+func GetTelnetBanner(logStruct *TelnetLog, conn net.Conn) error {
+	var err error
```
dadrian: If you're going to declare this on the first line, you may as well just make it a named return parameter.

` func GetTelnetBanner(logStruct *TelnetLog, conn net.Conn) err error { `

ztools/telnet/telnet.go
```diff
@@ -0,0 +1,94 @@
...
+	retBufferIndex := 0
+	var option, optionType byte
+	var iacIndex, prevIacIndex int
+	prevIacIndex = 0
+	for iacIndex = bytes.IndexByte(buffer, IAC); iacIndex != -1; iacIndex = bytes.IndexByte(buffer, IAC) {
+		optionType = buffer[iacIndex+1]
+		option = buffer[iacIndex+2]
+
+		if optionType == WILL || optionType == WONT {
+			optionType = DONT
```
zakird: Is this correct or backwards? (I don't know protocol, but seems like something that could be switched based on the naming)

ztools/telnet/telnet.go
```diff
@@ -0,0 +1,94 @@
...
+		return err
+	}
+
+	//grab banner
+	buffer := make([]byte, READ_BUFFER_LENGTH)
+
+	numBytes := len(buffer)
+	count := 0
+	for numBytes != 0 && count < 15 {
+		numBytes, err = conn.Read(buffer)
```
dadrian: You should also loop on `err != nil`

ztools/telnet/telnet.go
```diff
@@ -0,0 +1,94 @@
...
+	if err = NegotiateOptions(logStruct, conn); err != nil {
+		return err
+	}
+
+	//grab banner
+	buffer := make([]byte, READ_BUFFER_LENGTH)
+
+	numBytes := len(buffer)
+	count := 0
+	for numBytes != 0 && count < 15 {
```
dadrian: What's the point of count < 15? What end condition are your aiming for with this loop?

ztools/telnet/telnet.go
dadrian: Can you link the RFC for this?

dadrian:
Something seems off here, it doesn't currently work against ATT's public route view server (route-server.ip.att.net)

```bash
$ echo "12.0.1.28" | ./zgrab --port 23 --telnet | jq .
Nov 20 10:29:41.760 [ERROR] banner-grab: Conversation error with remote host 12.0.1.28:23: Unsupported telnet option type
{
  "ip": "12.0.1.28",
  "timestamp": "2015-11-20T10:29:41-05:00",
  "data": {
    "telnet": {}
  },
  "error": "Unsupported telnet option type",
  "error_component": "telnet"
}
```

zzma:
Sorry I'm on the road headed to NYC, I'll respond later tonight when I am free.




Zane

ztools/telnet/telnet.go
zzma: :100:

ztools/telnet/telnet.go
zzma: :+1:

ztools/telnet/telnet.go
zzma: I was artificially capping the length of the banner data response, but I will create a new flag telnet-max-size defaulting to a sane value that will determine how much of the telnet banner to read.

ztools/telnet/telnet.go
zzma: https://tools.ietf.org/html/rfc854 - page 14. 

Basically, the optionType is the telnet server indicating what it WILL or WONT support, and by default we are responding that the client (zgrab) requests the server DONT (or doesn't, in English) operate on a given option. When the telnet server sends a DONT or DO, it's asking the client (zgrab) to support or not support a given option, so by default we are responding that zgrab WONT support the option. 

DONT is zgrab telling the server not to do something. WONT is telling the server that he zgrab client will not do something. "something" here means a user-defined option code.

ztools/telnet/log.go
zzma: will do

zzma:
Closing this due to staleness and opening a new PR.

PR #93, opened 2015-12-02
Zzma/niagara fox banner grab

zakird:


zgrab_schema.py
```diff
@@ -393,6 +393,34 @@
 
 register_schema("zgrab-bacnet", zgrab_bacnet)
 
+zgrab_fox = Record({
+    "data": SubRecord({
+        "fox": SubRecord({
+            "is_fox": Boolean(),
+            "version": AnalyzedString(es_include_raw=True),
```
zzma: @zakird - any feedback on whether the appropriate strings are being analyzed?

ztools/fox/fox.go
```diff
@@ -0,0 +1,112 @@
...
+		return errors.New("Need a bigger buffer to read Fox response")
+	}
+	if err != nil {
+		return err
+	}
+
+	responseString := string(readBuffer[0:bytesRead])
+	output := strings.Split(responseString, string(0x0a))
+
+	if strings.HasPrefix(responseString, "fox a 0 -1 fox hello") {
```
zzma: This should use the RESPONSE_PREFIX constant defined above. I'll fix this

zgrab_schema.py
zzma: nvm see you already made some changes

zakird:
"hostname": "10.1.21.37",
     "host_address": "HVACSERVER15",
that seems switched?

zakird:
What do we name the subprotocol for this?

zzma:
Yes - it was switched. Good call. And I've seen primarily Fox protocol and sometimes Fox tunneling protocol. My vote is for Fox protocol

zakird:
Fox makes sense for protocol. I'm not sure what to do for subproto.

zakird:
I'd guess info or banner or device_id unless there's a command your'e running

zzma:
I don't know much about the protocol TBH, and a quick google doesn't shed any light, so I'm down to call it banner or device_id.

ztools/fox/fox.go
```diff
@@ -0,0 +1,112 @@
...
+import (
+	"encoding/hex"
+	"errors"
+	"net"
+	"strconv"
+	"strings"
+)
+
+const (
+	ORIGINAL_QUERY = "666f7820612031202d3120666f782068656c6c6f0a7b0a" +
```
dadrian: I'm not a huge fan of just putting in a query as binary, but it's not the worst thing.

ztools/fox/fox.go
```diff
@@ -0,0 +1,112 @@
...
+		"312c6f6e206f722061667465722c73756e6461792c756e646566696e6564" +
+		"0a686f737449643d733a57696e2d393943422d443439442d353434322d30" +
+		"3742420a766d557569643d733a38623533306263382d373663352d343133" +
+		"392d613265612d3066616264333934643330350a6272616e6449643d733a" +
+		"76796b6f6e0a7d3b3b0a"
+	RESPONSE_PREFIX = "fox a 0 -1 fox hello"
+)
+
+func GetFoxBanner(logStruct *FoxLog, connection net.Conn) error {
+	queryBytes, err := hex.DecodeString(ORIGINAL_QUERY)
```
dadrian: Move this to an `init()` block and make `queryBytes` be a global variable. It doesn't need to run on every host.

ztools/fox/fox.go
zzma: agreed. but I don't know the full details of the protocol and haven't had the time to fully parse through everything. this is mostly ported from nmap

dadrian:
Looks pretty good overall.

PR #100, opened 2015-12-24
Zzma/telnet banner grab

zzma:
@dadrian @zakird - Basic telnet banner grab.

The erroneous cases that were occurring before were due to multiple rounds of DO/DONT/WILL/WONT option negotiation, and also a slightly too small default banner read size.

Tested on a handful of telnet servers (including AT&T 12.0.1.28) and looks good.

### DIFF
--- a/ztools/telnet/telnet.go
+++ b/ztools/telnet/telnet.go
@@ -40,6 +40,7 @@ func GetTelnetBanner(logStruct *TelnetLog, conn net.Conn, maxReadSize int) (err 
 		return err
 	}
 
+	var bannerSlice []byte
 	//grab banner
 	buffer := make([]byte, READ_BUFFER_LENGTH)
 
@@ -67,12 +68,14 @@ func GetTelnetBanner(logStruct *TelnetLog, conn net.Conn, maxReadSize int) (err 
 		}
 
 		if count == rounds-1 {
-			logStruct.Banner = logStruct.Banner + string(buffer[0:maxReadSize%READ_BUFFER_LENGTH])
+			bannerSlice = append(bannerSlice, buffer[0:maxReadSize%READ_BUFFER_LENGTH]...)
 		} else {
-			logStruct.Banner = logStruct.Banner + string(buffer[0:numBytes])
+			bannerSlice = append(bannerSlice, buffer[0:numBytes]...)
 		}
 		count += 1
 	}
+
+	logStruct.Banner += string(bannerSlice)
 
 	return nil
 }
--- a/ztools/telnet/telnet.go
+++ b/ztools/telnet/telnet.go
@@ -16,46 +16,56 @@ import (
 )
 
 var (
-	IAC            = 0xFF //255 - Interpret as command
-	DONT           = 0xFE
-	DO             = 0xFD
-	WONT           = 0xFC
-	WILL           = 0xFB
-	READ_TIMEOUT   = 3 * time.Second
-	IAC_CMD_LENGTH = 3 // IAC commands take 3 bytes (inclusive)
+	IAC                = 0xFF //255 - Interpret as command
+	DONT               = 0xFE
+	DO                 = 0xFD
+	WONT               = 0xFC
+	WILL               = 0xFB
+	READ_TIMEOUT       = 3 * time.Second
+	IAC_CMD_LENGTH     = 3 // IAC commands take 3 bytes (inclusive)
+	READ_BUFFER_LENGTH = 8192
 )
 
 func GetTelnetBanner(logStruct *TelnetLog, conn net.Conn) error {
+	conn.SetReadDeadline(time.Now().Add(READ_TIMEOUT))
 
 	if err := NegotiateOptions(conn); err != nil {
 		return err
 	}
 
 	//grab banner
-
-	return nil
-
-}
-
-func NegotiateOptions(conn net.Conn) error {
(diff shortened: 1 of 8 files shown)


zakird:
Offhand this looks good to me. Have you spot checked at all with telnet
client to make sure we're not missing any hosts?

Zakir

zzma:
I've tested some know telnet servers, and also done some 1% testing. Do you mean passing zmap output into the native Linux telnet client and checking error codes or something?

zakird:
Yeah. Just spot check a handful and make sure there aren't ones that work
with the telnet client that don't work with ours

ztools/telnet/telnet.go
```diff
@@ -0,0 +1,137 @@
...
+	"bytes"
+	"errors"
+	"fmt"
+	"io"
+	"math"
+	"net"
+)
+
+const (
+	IAC                = byte(255) //Interpret as command
```
dadrian: Can you link to the RFC in a comment here, just so we can remember where these values came from in the future?

zzma:
Tested 50 random telnet IP addresses, and there seems to be parity between native telnet and zgrab! 

The one caveat to this is that there are cases where the banner is completely read, but there is a i/o connection timeout. This occurs because there is no formal "end-of-data" command/marker in telnet, so the telnet client must continually attempt to read from the server. In some cases, the server immediately yields an empty response, which we interpret as a lack of additional data and stop reading. In other cases, the server hangs on the connection and this causes the i/o timeouts that we observe.

I'm wondering if we should do special error handling and only report i/o timeouts if there is no telnet banner that has been read thus far. If there is already a banner that has been received from a previous Read, then we assume that everything worked as expected and squelch the error.

zakird:
Awesome! Re second case, I agree what you suggested makes sense. Can we set
a minimum length and if it that's met, we call that good? How often does
this happen?

dadrian:
How do we detect the immediate read of zero content? Do we get an io.EOF?

dadrian:
Also, do you have a couple example outputs?

ztools/telnet/telnet.go
```diff
@@ -0,0 +1,137 @@
...
+import (
+	"bytes"
+	"errors"
+	"fmt"
+	"io"
+	"math"
+	"net"
+)
+
+// RFC 854 - https://tools.ietf.org/html/rfc854
```
dadrian: Yay!

ztools/telnet/telnet.go
```diff
@@ -0,0 +1,137 @@
...
+		return err
+	}
+
+	//grab banner
+	buffer := make([]byte, READ_BUFFER_LENGTH)
+
+	numBytes := len(buffer)
+	rounds := int(math.Ceil(float64(maxReadSize) / READ_BUFFER_LENGTH))
+	count := 0
+	for numBytes != 0 && count < rounds && numBytes == READ_BUFFER_LENGTH {
```
dadrian: I'm confused by this loop. Why not just attempt to read `maxReadSize` bytes no more than `READ_BUFFER_LENGTH / max_read_size + 1` times? Also, why limit number of rounds if you're not erroring? Can we just attempt to read up to `READ_BUFFER_LENGTH` bytes or until an error is encountered?

zzma:
@dadrian - yes we detect a read of zero-content with either io.EOF or numBytes == 0

Here's some sample output: 
```
{"ip":"162.201.103.209","timestamp":"2015-12-29T00:54:23-06:00","data":{"telnet":{"banner":"\r\n\r\nUser Access Verification\r\n\r\nPassword: ","supported_opts":["1","3","31"]}}}
{"port":23,"success_count":1,"failure_count":0,"total":1,"start_time":"2015-12-29T00:54:23-06:00","end_time":"2015-12-29T00:54:24-06:00","duration":0,"senders":1000,"timeout":10,"tls_version":null,"mail_type":null,"ca_file_name":null,"sni_support":true}
```

ztools/telnet/telnet.go
zzma: I think we have flipped understandings of maxReadSize and READ_BUFFER_LENGTH. maxReadSize is the total number of bytes that are configured to be read, and READ_BUFFER_LENGTH is the length of the buffer passed to `conn.Read`. The number of rounds is the way that attempt to read READ_BUFFER_LENGTH bytes no more than `max_read_size / READ_BUFFER_LENGTH  + 1` times (except in the case of the last round, where we only want to read the remainder `max_read_size % READ_BUFFER_LENGTH` number of bytes) . The check for numBytes != 0 is to not process the banner if there is no data that has been read (which can happen even without a io.EOF err), and we also consider `conn.Read` to be finished if it doesn't fill up the full READ_BUFFER_LENGTH (i.e. only a partial buffer's worth of data has been returned by the server)

zzma:
@zakird - 1291 out of 3024 telnet hosts with banners timed-out. I've gone ahead and safely handled the error when there is any banner data present.

ztools/telnet/telnet.go
dadrian: Ah, I did reverse the meaning of those variables. In that case we should use `append()` to join the buffers populated by `conn.Read()`, and then convert to a string outside of the loop.

zzma: done-so.

dadrian:
:shipit:

PR #104, opened 2016-01-04
Add ability to gather TLS Session Tickets.

aaspring:
The --tls-session-ticket-ext flag adds the Session Ticket extension
support to TLS Client Hello. If the server sends a New Session Ticket
payload, then the ticket value is extracted and saved in JSON output.

### DIFF
--- a/zlib/grabber.go
+++ b/zlib/grabber.go
@@ -19,16 +19,16 @@ import (
 	"encoding/csv"
 	"errors"
 	"fmt"
-	"io"
-	"net"
-	"strconv"
-	"time"
 	"github.com/zmap/zgrab/ztools/ftp"
 	"github.com/zmap/zgrab/ztools/processing"
 	"github.com/zmap/zgrab/ztools/scada/dnp3"
 	"github.com/zmap/zgrab/ztools/scada/fox"
 	"github.com/zmap/zgrab/ztools/scada/siemens"
 	"github.com/zmap/zgrab/ztools/telnet"
+	"io"
+	"net"
+	"strconv"
+	"time"
 )
 
 type GrabTarget struct {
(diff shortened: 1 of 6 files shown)


zakird:
Tests currently fail for this request. Can you please correct any CI issues?

zlib/grabber.go
```diff
@@ -19,16 +19,16 @@ import (
...
 	"github.com/zmap/zgrab/ztools/ftp"
 	"github.com/zmap/zgrab/ztools/processing"
 	"github.com/zmap/zgrab/ztools/scada/dnp3"
 	"github.com/zmap/zgrab/ztools/scada/fox"
 	"github.com/zmap/zgrab/ztools/scada/siemens"
 	"github.com/zmap/zgrab/ztools/telnet"
+	"io"
+	"net"
+	"strconv"
+	"time"
```
zakird: Why did this order change?

dadrian:
The tests are failing because the TLS library (by default), sends session tickets, and the new code does not. See https://golang.org/src/crypto/tls/common.go?s=8571:13264#L235

```
type Config struct {
...
    // SessionTicketsDisabled may be set to true to disable session ticket
    // (resumption) support.
    SessionTicketsDisabled bool
...
}
```

zakird:
If golang does this by default, we probably don't need to add an additional flag.

aaspring:
As-is, ZGrab does not send TLS Session Ticket support extension when passing the --tls flag. This pull request is maintaining that behavior hence the need for a new flag. Currently trouble shooting with @dadrian

zlib/grabber.go
aaspring: gofmt sorts the imports alphabetically

PR #106, opened 2016-01-06
Extract and output session ticket lifetime hint

aaspring:
Extract the ticket_lifetime_hint field from the New Session Ticket payload and add to JSON output.


zgrab_schema.py
```diff
@@ -202,7 +202,8 @@
     }),
     "session_ticket":SubRecord({
         "value":Binary(),
-        "length":Integer()
+        "length":Integer(),
```
dadrian: What's the point of having length? Length is implicitly determined by the base64 value stored in the `value` field in JSON.

zakird: It's like how we store length for every other binary thing. It's difficult
to compute this in place and varies.

aaspring: Was requested by @zakird due to ``it's a pain to calculate that on the fly''. Length is not that of base64 string, but of the ticket encoded in base64.

dadrian: We usually only store length (in bits) for primes and keys. We've removed it in other places, such as the ZMap JSON output.

zakird: I think this is determined more by "does it vary wildly" between hosts. For
something like keys, it does. For other things, where it's always the same,
it doesn't matter. So, I would argue that we have it if it varies.

dadrian: If it's important (e.g. some sort of security parameter), I'm fine storing length. It sounds like in this case it is. @aaspring, what's the length range, and do you think it makes more sense to give length in bits or bytes?

aaspring: Length range according to RFC 5077 is 0 -> 2^16 - 1 but don't know what the real-world lengths are (hence the ZGrab patch). Definitely makes sense to be in bytes.

dadrian: Ok. If we're planning on doing analysis based on lengths, I'm fine leaving it and we can merge.

ztools/ztls/handshake_messages.go
```diff
@@ -1394,6 +1395,7 @@ func (m *newSessionTicketMsg) unmarshal(data []byte) bool {
 		return false
 	}
 
+	m.lifetimeHint = uint32(data[4])<<24 | uint32(data[5])<<16 | uint32(data[6])<<8 | uint32(data[7])
```
zakird: Is there a prettier way of doing this in golang? /cc @dadrian

zakird: https://golang.org/pkg/encoding/binary/ see read functions

dadrian: There's a binary module, but I've seen it done both ways. Usually using the binary package isn't worth it unless you're doing a lot of marshaling. In this case, I don't think it matters and we're fine building our own int from bytes directly.

dadrian: https://golang.org/pkg/encoding/binary/

zakird:
alright, this lgtm.

PR #107, opened 2016-01-08
Add RFC 7627 Extended Master Secret extension

aaspring:
Add extended master secret TLS extension through the use of the ``--tls-extended-master-secret'' command line flag. Tested against google.com's implementation IOT ensure accuracy in computation of master secret and ability to continue interaction once TLS handshake is complete.


dadrian:
I think this is implemented in either the newer Golang stdlibrary, or the BoringSSL test runner. We should just use that.

aaspring:
Closing PR, new PR inbound shortly patching in BoringSSL test/runner implementation.

PR #108, opened 2016-01-09
Add RFC 7627 Extended Master Secret extension

aaspring:
Add extended master secret TLS extension through the use of the
``--tls-extended-master-secret'' command line flag. Tested against
google.com's implementation IOT ensure accuracy in computation of master
secret and ability to continue interaction once TLS handshake is
complete.

Additions are ported from BoringSSL
(https://boringssl.googlesource.com/boringssl
hash 7b8b9c17db93ea5287575b437c77fb36eeb81b31) as close as possible.
Only main difference is that this commit uses the
Config.ExtendedMasterSecret flag to enable or disable support for
extension (disabled without command-line flag).


ztools/ztls/handshake_client.go
```diff
@@ -580,7 +588,6 @@ func (hs *clientHandshakeState) doFullHandshake() error {
 		sr = hs.serverHello.random
 	}
 
-	hs.masterSecret = masterFromPreMasterSecret(c.vers, hs.suite, preMasterSecret, cr, sr)
```
dadrian: The master secret calculation needs to be moved here to account for extended random.

dadrian:
Does BoringSSL have any test for extended master secret that we could port over?

zakird:
Did we end up merging in BoringSSL here in the end? Can this pull request be closed?

PR #112, opened 2016-02-24
Add --output-crypto-variables flag.

aaspring:
Cmd flag adds premaster secret and master secret to the JSON output of
the ZGrab targets IOT avoid re-computation in post-processing and
analysis.

### DIFF
--- a/main.go
+++ b/main.go
@@ -112,7 +112,7 @@ func init() {
 
 	flag.BoolVar(&config.GatherSessionTicket, "tls-session-ticket", false, "Send support for TLS Session Tickets and output ticket if presented")
 	flag.BoolVar(&config.ExtendedMasterSecret, "tls-extended-master-secret", false, "Offer RFC 7627 Extended Master Secret extension")
-	flag.BoolVar(&config.TLSVerbose, "tls-verbose", false, "Add client_hello and key_material to JSON output")
+	flag.BoolVar(&config.TLSVerbose, "tls-verbose", false, "Add extra TLS information to JSON output (client hello, client KEX, key material, etc)")
 
 	flag.StringVar(&rootCAFileName, "ca-file", "", "List of trusted root certificate authorities in PEM format")
 	flag.IntVar(&config.GOMAXPROCS, "gomaxprocs", 3, "Set GOMAXPROCS (default 3)")
--- a/main.go
+++ b/main.go
@@ -112,6 +112,7 @@ func init() {
 
 	flag.BoolVar(&config.GatherSessionTicket, "tls-session-ticket", false, "Send support for TLS Session Tickets and output ticket if presented")
 	flag.BoolVar(&config.ExtendedMasterSecret, "tls-extended-master-secret", false, "Offer RFC 7627 Extended Master Secret extension")
+	flag.BoolVar(&config.OutputCryptoVariables, "output-crypto-variables", false, "Explicitly add crypto variables (PMS & MS) to JSON output")
 
 	flag.StringVar(&rootCAFileName, "ca-file", "", "List of trusted root certificate authorities in PEM format")
 	flag.IntVar(&config.GOMAXPROCS, "gomaxprocs", 3, "Set GOMAXPROCS (default 3)")
--- a/main.go
+++ b/main.go
@@ -112,7 +112,7 @@ func init() {
 
 	flag.BoolVar(&config.GatherSessionTicket, "tls-session-ticket", false, "Send support for TLS Session Tickets and output ticket if presented")
 	flag.BoolVar(&config.ExtendedMasterSecret, "tls-extended-master-secret", false, "Offer RFC 7627 Extended Master Secret extension")
-	flag.BoolVar(&config.LogMasterSecret, "log-master-secret", false, "Add premaster and master secrets to JSON output")
+	flag.BoolVar(&config.TLSVerbose, "tls-verbose", false, "Add client_hello and key_material to JSON output")
 
 	flag.StringVar(&rootCAFileName, "ca-file", "", "List of trusted root certificate authorities in PEM format")
 	flag.IntVar(&config.GOMAXPROCS, "gomaxprocs", 3, "Set GOMAXPROCS (default 3)")
@@ -127,6 +127,7 @@ func init() {
 	flag.BoolVar(&config.SSH.NegativeOne, "ssh-negative-one", false, "Set SSH DH kex value to -1 in the selected group")
 	flag.BoolVar(&config.Telnet, "telnet", false, "Read telnet banners")
 	flag.IntVar(&config.TelnetMaxSize, "telnet-max-size", 65536, "Max bytes to read for telnet banner")
+
 	flag.Parse()
 
 	// Validate Go Runtime config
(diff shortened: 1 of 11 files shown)


main.go
```diff
@@ -112,6 +112,7 @@ func init() {
 
 	flag.BoolVar(&config.GatherSessionTicket, "tls-session-ticket", false, "Send support for TLS Session Tickets and output ticket if presented")
 	flag.BoolVar(&config.ExtendedMasterSecret, "tls-extended-master-secret", false, "Offer RFC 7627 Extended Master Secret extension")
+	flag.BoolVar(&config.OutputCryptoVariables, "output-crypto-variables", false, "Explicitly add crypto variables (PMS & MS) to JSON output")
```
zakird: Let's name this something more descriptive like --out-master-secret

aaspring: @dadrian and I previously discussed possibly adding this to the default output but no final decision was made. Is that still a possibility or are we going w/ "--out-master-secret"?

zakird: I'm not thrilled about the idea if we're not going to use the data. That will add gigs work of data to every HTTPS scan.

zgrab_schema.py
```diff
@@ -205,7 +205,11 @@
         "value":Binary(),
         "length":Integer(),
         "lifetime_hint":Long()
-    })
+    }),
+    "crypto_variables":SubRecord({
```
zakird: I'm not convinced this is the most descriptive thing to call this, or place to put these variables. Practically everything in the TLS handshake is a crypto variable. pre-master-secret should go under client_key_exchange. Not totally sure where we want to put master_secret.

aaspring: There is no current client_key_exchange JSON output record for TLS. The only related output record is the server_key_exchange which contains the server's public values and signature. I can create a new client_key_exchange JSON record including the client's public and private values if needed. Putting the premaster secret there kinda makes sense, but not completely b/c it is dependent on both the server KEX and client KEX records' values.

The same goes with the master secret. It kinda makes sense to put it in the server_key_exchange record, but not completely b/c it's not wholly based on the server KEX. I could put it in the server_finished b/c that's the point in the handshake that it's required, but it doesn't make logical sense.

That's why I created a new data->tls layer record to hold these values. They're dependent on multiple parts of the existing JSON record structure and don't fit well WRT picking one to insert it into or the common-sense lay out of the existing output.

zakird: I'm alright putting it as a top level thing if we can find a descriptive name for it.

dadrian:
Alright, so after some talking with @zakird, we came to a couple conclusions that I think will make everybody happy and also solve several problems.

We add a pointer to the `KeyMaterial` struct described in this Gist (https://gist.github.com/dadrian/f655c90ce3993d828655) to the `ServerHandshake` struct. This contains the pre-master secret and the master secret, and is structured such that we can easily add new fields.

- We add a flag `--tls-verbose` that defaults to false.
- When `--tls-verbose` is true:
    - We output the `client_hello` and `key_material` fields only when this flag is passed.

This is slightly different behavior than we have now, in that outputting `client_hello` will no longer be default. That being said, I don't think anyone ever uses the client hello except for Red Hat, once, last summer.

@aaspring, does this work for you?

aaspring:
That works for me.

aaspring:
While we're at it, it may be worth thinking about adding a client_key_exchange and client_finished group to the --tls-verbose behavior.

dadrian:
@aaspring I definitely agree those should be controlled by this flag as well, but I wasn't going to make you implement that now unless you really wanted to.

dadrian:
Looks good overall. We could maybe add a `length` field for each of the secrets, but I'm not sure if that's necessary or not.

zgrab_schema.py
```diff
@@ -205,7 +205,64 @@
...
+                "x":SubRecord({
+                    "value":Binary(),
+                    "length":Integer()
+                }),
+                "y":SubRecord({
+                    "value":Binary(),
+                    "length":Integer()
+                }),
+            }),
+            "client_private":SubRecord({
```
dadrian: Is this also not X and Y?

zgrab_schema.py
```diff
@@ -205,7 +205,64 @@
...
+                "value":Binary(),
+                "length":Integer()
+            }),
+        }),
+        "rsa_params":SubRecord({
+            "length":Integer(),
+            "encrypted_pre_master_secret":Binary()
```
dadrian: This will be the decrypted version, no?

zgrab_schema.py
aaspring: This is the client secret used to turn G into X and Y in the client_public.

dadrian: OK

zgrab_schema.py
aaspring: No, this is the encrypted version sent over the wire to the server. The unencrypted premaster secret is included in the key_material record and removed here IOT reduce duplication. With the --tls-verbose flag, you can't output the client KEX record w/o the key material record.

```golang
func (m *clientKeyExchangeMsg) MakeLog(ka keyAgreement) *ClientKeyExchange {
 	ckx := new(ClientKeyExchange)
 	ckx.Raw = make([]byte, len(m.raw))
 	copy(ckx.Raw, m.raw)
 
 	switch ka := ka.(type) {
 	case *rsaKeyAgreement:
 		ckx.RSAParams = new(keys.RSAClientParams)
 		ckx.RSAParams.Length = uint16(len(m.ciphertext) - 2) // First 2 bytes are length
 		ckx.RSAParams.EncryptedPMS = make([]byte, len(m.ciphertext)-2)
 		copy(ckx.RSAParams.EncryptedPMS, m.ciphertext[2:])
 		// Premaster-Secret is available in KeyMaterial record
 	case *dheKeyAgreement:
 		ckx.DHParams = ka.ClientDHParams()
 	case *ecdheKeyAgreement:
 		ckx.ECDHParams = ka.ClientECDHParams()
 	default:
 		break
 	}
 
 	return ckx
 }
```

dadrian: Do we actually need to output the encrypted one then? I guess we may as well?

aaspring: It keeps the mental model of the JSON output consistent. Otherwise random JSON output records would be completely missing a client_key_exchange sub-record b/c the server selected Client-Write and there is no other information in the client KEX.

dadrian:
Looks good to me. I'm good with merge. @zakird, any concerns?

zakird:
I'm fundamentally fine with this. However, I'm not a big fan of the terms `client_public` and `client_private`. Do I understand correctly that these are g^x (mod p) and x? Can we just name them what they are? Do we need to have client_public/ x^x given that this is trivially calculable?

aaspring:
The ```client_public``` and ```client_private``` terms were used to duplicate and stay logically consistent with the structure of the ```server_public``` JSON record.

Yes, you could re-calculate in post-processing, but this PR is intended make as much as possible easily available to analysis scripts. Instead of the processing script having to do ECC itself IOT have the client's X and Y, it can read it directly from the output. As discussed previously, the additional size is of little concern as this is not a default behavior.

zakird:
I'm fine with keeping it. @dadrian what are your thoughts on naming?

dadrian:
I was opposed until I saw that the naming matched existing output for server.

zakird:
fwm

dadrian:
Merged in 7ae7f9b69adc0230efa3d3e58d6885e2eb530832

PR #116, opened 2016-03-11
Toward X.509 certificates that json.Unmarshal

dadrian:
This is to track progress towards implementing the missing json.Unmarshal functions in the X.509 library, which is required to have a decent implemention of json.Unmarshal within ztls.

This should not be merged yet.


zakird:
@dadrian this PR cannot be merged into master and has been sitting open for several months now. Do you plan on fixing this problems in the near future (i.e., the next week)? If not, let's close this and we can resubmit a PR when it's ready to be landed. I will close this PR on 6/27 if I don't hear otherwise from you.

Issue #117, opened 2016-03-20
ZGrab incorrectly outputs TLS Session Ticket support as false


zakird:
If we do not offer TLS Session Ticket support, we output that the server does not support the extensions, which is not strictly correct. We should not output this, or output it as _null_ , if we haven't tested for it. This has bitten us in the past for other fields.

aaspring:
Identical behavior is present in ```extended_master_secret``` output field and should be fixed in the same PR. Attaching to this Issue IOT for tracking.

zakird:
@dadrian, can you take a look at this when you have a moment?

PR #119, opened 2016-03-23
Fix confusing JSON output in server_hello.

aaspring:
* Fixes Issue #117
    * Extended master secret and session ticket ext show as 'not
        supported' if not offered
* Patch is overly complicated compared to others b/c empty/default
    boolean value is false
    * Using 'omitempty' would mean non-supporting servers would not
        have field when offered

### DIFF
--- a/ztools/ztls/handshake_client.go
+++ b/ztools/ztls/handshake_client.go
@@ -165,7 +165,7 @@ func (c *Conn) clientHandshake() error {
 		c.sendAlert(alertUnexpectedMessage)
 		return unexpectedMessageError(serverHello, msg)
 	}
-	c.handshakeLog.ServerHello = serverHello.MakeLog()
+	c.handshakeLog.ServerHello = serverHello.MakeLog(c.config)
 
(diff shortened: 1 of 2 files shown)


dadrian:
There's a cleaner way to do this. Instead of manually constructing the JSON string, in the `MarshallingServerHello`, add session_ticket as a `*bool` set to `omitempty`. Then the `nil` pointer will be omitted, but a pointer to true or false will be printed. You'll need to set the appropriate value for the pointer in the `MarshalJSON` method.

dadrian:
@aaspring do you want to fix, or should I?

dadrian:
@aaspring bump <3

zakird:
This pull request has been sitting open for almost two months now. What's it's status? 
@aaspring do you plan on fixing the problems that @dadrian mentioned?

I'm going to close this PR on 6/27 if I don't hear otherwise from one of the two you.

zakird:
Closing due to lack of progress or response.

PR #120, opened 2016-03-26
Length-limit headers to < 8K

dadrian:
@zzma, thoughts?

### DIFF
--- a/ztools/http/header.go
+++ b/ztools/http/header.go
@@ -87,19 +87,30 @@ type UnknownHeader struct {
 	Values []string `json:"value,omitempty"`
 }
 
+func formatHeaderValues(v []string) {
+	for idx := range v {
+		if len(v[idx]) >= 8192 {
+			v[idx] = v[idx][0:8191]
+		}
+	}
+}
+
 // Custom JSON Marshaller to comply with snake_case header names
 func (h Header) MarshalJSON() ([]byte, error) {
 	headerMap := make(map[string]interface{})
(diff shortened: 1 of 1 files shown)


zakird:
I think we should do this for both known and unknown headers.

zzma:
@zakird - it looks like it does do for both known and unknown

zzma:
@dadrian - looks good to me

zzma:
curious - were we running into issues due to unconstrained header length?

zakird:
`Mar 26 19:13:38.356 [INFO] esloader: unknown elastic search error code 500, {u'index': {u'status': 500, u'_type': u'host', u'_id': u'1755048135', u'_index': u'ipv4', u'error': u'IllegalArgumentException[Document contains at least one immense term in field="80.http.get.headers.unknown.value.raw" (whose UTF8 encoding is longer than the max length 32766), all of which were skipped. Please correct the analyzer to not produce such terms. The prefix of the first immense term is: \'[101, 121, 74, 50, 90, 88, 74, 122, 97, 87, 57, 117, 73, 106, 111, 105, 78, 67, 52, 120, 76, 106, 65, 105, 76, 67, 74, 106, 98, 50]...\', original message: bytes can be at most 32766 in length; got 121484]; nested: MaxBytesLengthExceededException[bytes can be at most 32766 in length; got 121484]; '}}`

zakird:
@zzma good call, missed the second invocation. lgtm.

PR #121, opened 2016-03-27
add pdu length field to s7 request

zzma:
@zakird 
FYI - going to go ahead and auto-merge. While re-visiting S7 protocol, I noticed that adding this field increases the number of detected S7 devices without losing any devices that are already  being detected.


zakird:
lgtm.

PR #122, opened 2016-04-01
Remove extraneous log in HTTP module

dadrian:
/cc @zzma @zakird


zzma:
Looks good to me. Where were you seeing this pop up?

zzma:
Looks like there's an extraneous "log" package inclusion causing the test failure.

PR #126, opened 2016-04-27
don't hold on to idle connections

zzma:
@zakird 

This seems to fix the main memory leak error and hopefully will resolve the issues we were seeing for the HTTP scans. Basically, we were caching every connection and never purging. The other thing is that we never really utilized the cache, so this change should not noticeably affect performance.


dadrian:
❤️

PR #127, opened 2016-04-27
Zzma/handle only domain inputs

zzma:
@zakird @dadrian 

Adding a new command line option "--lookup-domain" that expects a list of domain names and will look them up. Currently only works for HTTP, but can be expanded to other protocols, if necessary. This is the stuff that Ben is looking for. 

Question - should I make the name for this option have an "http" prefix? Do we envision doing DNS lookups for other domains?

### DIFF
--- a/zlib/grabber.go
+++ b/zlib/grabber.go
@@ -15,6 +15,7 @@
 package zlib
 
 import (
+	"bufio"
 	"bytes"
 	"crypto/sha256"
 	"encoding/csv"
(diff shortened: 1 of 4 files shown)


zakird:
Let's wait to merge this until we can do this generically for every protocol. Ben actually needs this for HTTPS. Unless this is going to be very fast, let's put this on the back burner until after CCS is submitted. It sounds like Ben has an interim solution. I would like to get merged at some point though. This seems like useful functionality for everyone.

zzma:
Note to self: stress test this.

zakird:
This has been sitting open for quite a while now. @zzma is this still what we want to merge?

@dadrian Can you please see if you have any concerns about this PR by 6/27? If not, we'll plan to merge it.

zlib/grabber.go
```diff
@@ -121,9 +150,6 @@ func makeHTTPGrabber(config *Config, grabData GrabData) func(string) error {
 			tlsConfig.RootCAs = config.RootCAPool
 			tlsConfig.HeartbeatEnabled = true
 			tlsConfig.ClientDSAEnabled = true
-			//		if !config.NoSNI && session.domain != "" {
-			//			tlsConfig.ServerName = session.domain
-			//		}
```
dadrian: Wait a minute, why was this commented out in the first place?

zzma: I had originally commented out the entire section when first writing the code, but forgot to add it back in. This has already been changed in master and should go away once I rebase.

zzma:
The addition of the `--lookup-domain` flag allows us to process only domain inputs (with no IP address). It throws an error when trying to parse IP, domain CSV. Do you guys think renaming to `--domain-only` would be clearer?

Unrelated, but I discovered that when an IP, domain are input as CSV, we leave the http host as the ip address. Should we change/fix this so that the host is set as the provided domain, while still making the connection to the provided IP address?

zakird:
We should absolutely set it to the domain not the IP address.

zakird:
I'm fine with leaving it ``--lookup-domain``. That makes sense for now and we can consider long term when we move things around for ZGrab 2.0.

zzma:
@zakird - I think this is ready to merge. I will add the feature to set the domain and not the IP address (when both are provided) as the Host in a subsequent PR

PR #128, opened 2016-05-10
actually do https when --tls flag is specified

zzma:
@zakird @dadrian 

Do HTTPS when --TLS flag is specified. I must have overlooked this when implementing the new HTTP stuff.


zzma:
For the record, the `--tls` flag on its own was not broken. Only the `--tls` flag in conjunction with the `--http` flag

PR #131, opened 2016-06-03
Clean up cipher suite settings

corny:
I need to set the cipher suites in a `ztls.Conn` directly.
This refactoring cleans up the code and intruduces the `CipherSuites` field.

### DIFF
--- a/zlib/conn.go
+++ b/zlib/conn.go
@@ -64,16 +64,8 @@ type Conn struct {
 
 	caPool *x509.CertPool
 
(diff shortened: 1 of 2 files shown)


dadrian:
This looks good to me. Our old abstraction was awful anyway. Have you tested all flags?

corny:
I only tested it by setting lists of cipher suites directly. It compiles and this works for me.

corny:
Ready to merge?

PR #135, opened 2016-06-14
SCT bound checks

zakird:


zakird:
@shoegaze can you glance at and see if you have any concerns? If not, we can merge.

shoegaze:
Looks good to me.

PR #155, opened 2016-07-09
Zzma/set http user agent

zzma:
@zakird @dadrian 

Proper HTTP User Agent handling. The defaults and command line custom user agent don't currently work. This fixes that.


zzma:
Bump. This is ready to roll 🎢

PR #157, opened 2016-07-10
Certparty

### LINKED ISSUE #140: Add names field
collect all names that the certificate is valid for and put them in one place

shoegaze:
Closed #140.


ztools/x509/json.go
```diff
@@ -202,8 +201,10 @@ type jsonCertificate struct {
 	FingerprintMD5            CertificateFingerprint       `json:"fingerprint_md5"`
 	FingerprintSHA1           CertificateFingerprint       `json:"fingerprint_sha1"`
 	FingerprintSHA256         CertificateFingerprint       `json:"fingerprint_sha256"`
+	SPKIFingerprint           CertificateFingerprint       `json:"spki_fingerprint"`
```
dadrian: Wut.

ztools/x509/json.go
```diff
@@ -167,11 +167,10 @@ func (v *validity) UnmarshalJSON(b []byte) error {
...
 type jsonSubjectKeyInfo struct {
-	KeyAlgorithm    PublicKeyAlgorithm     `json:"key_algorithm"`
-	RSAPublicKey    *keys.RSAPublicKey     `json:"rsa_public_key,omitempty"`
-	DSAPublicKey    interface{}            `json:"dsa_public_key,omitempty"`
-	ECDSAPublicKey  interface{}            `json:"ecdsa_public_key,omitempty"`
-	SPKIFingerprint CertificateFingerprint `json:"fingerprint_sha256"`
+	KeyAlgorithm   PublicKeyAlgorithm `json:"key_algorithm"`
+	RSAPublicKey   *keys.RSAPublicKey `json:"rsa_public_key,omitempty"`
+	DSAPublicKey   interface{}        `json:"dsa_public_key,omitempty"`
+	ECDSAPublicKey interface{}        `json:"ecdsa_public_key,omitempty"`
```
dadrian: Wut.

ztools/x509/json.go
```diff
@@ -220,7 +221,16 @@ func (c *Certificate) MarshalJSON() ([]byte, error) {
 	jc.Subject = c.Subject
 	jc.SubjectDN = c.Subject.String()
 	jc.SubjectKeyInfo.KeyAlgorithm = c.PublicKeyAlgorithm
-	jc.SubjectKeyInfo.SPKIFingerprint = c.SPKIFingerprint
+
+	// 1 for the DN, and then however many DNS names there are
+	jc.Names = make([]string, 1+len(c.DNSNames))
+
+	jc.Names[0] = jc.SubjectDN
```
dadrian: The `SubjectDN` isn't a URL. I think you'll need to go through each of `[]Names` in `pkix.Name`, check to see if it's a URL, and add it if it is. Golang's `net.URL` probably helps with this.

ztools/x509/json.go
shoegaze: You see, I asked @zakird to check my sanity...

ztools/x509/json.go
shoegaze: Ah. My brain interpreted "DN" as domain name, not distinguished name.

shoegaze: Would it make more sense to reconstruct them from the domain component?

ztools/x509/json.go
```diff
@@ -221,7 +223,25 @@ func (c *Certificate) MarshalJSON() ([]byte, error) {
...
+			}
+		}
+	}
+
+	for _, name := range c.DNSNames {
+
+		if _, err := url.Parse(name); err == nil {
+			jc.Names = append(jc.Names, name)
+		}
+	}
```
dadrian: What happens when one of these contains `*.example.com`?

shoegaze: https://play.golang.org/p/kV7FopPrh0

shoegaze: Actually, as it turns out net/url is the wrong library to validate urls. Switching to govalidate, see here. https://www.socketloop.com/tutorials/golang-how-to-validate-url-the-right-way

shoegaze: Grr. Except this doesn't work either, apparently.

shoegaze:
bumping this, I think it's ready to go.

ztools/x509/json.go
```diff
@@ -221,7 +224,33 @@ func (c *Certificate) MarshalJSON() ([]byte, error) {
...
+				jc.Names = append(jc.Names, name)
+			}
+		}
+	}
+
+	for _, name := range c.DNSNames {
+
+		if name[0] == '*' {
+			name = name[2:]
+		}
```
zakird: Do just lose the wildcard here then? Shouldn't something like `*.zmap.io` be what's added?

shoegaze: ah, good catch. Fixing.

zakird:
We should also add this to the zgrab and ztag schemas.

shoegaze:
I'm working on adding more test data. I'll get back to you.

shoegaze:
Okay, now I have tested more thoroughly.

zakird:
Sounds good. I'm with merging once the field is in the schema.

ztools/x509/json.go
```diff
@@ -33,31 +35,31 @@ type auxKeyUsage struct {
 func (k KeyUsage) MarshalJSON() ([]byte, error) {
 	var enc auxKeyUsage
 	enc.Value = uint32(k)
-	if k & KeyUsageDigitalSignature > 0 {
+	if k&KeyUsageDigitalSignature > 0 {
```
zzma: Nitpick: we should not undo the proper formatting here

shoegaze: This was done by go fmt

zzma: Oh true. That's weird, looks like my editor auto-formats things differently from `go fmt`. I'll fix my setup

Issue #147, opened 2016-06-25
Parse out EV and OV


zakird:

zakird:
Thus, under the Baseline Requirements, additional OIDs adopted by the CA / Browser Forum are:
domain-validated(1) => (2.23.140.1.2.1)  (Compliant with Baseline Requirements – No entity identity asserted)
organization-validated(2) => (2.23.140.1.2.2)  (Compliant with Baseline Requirements – Organization identity asserted)
individual-validated(3) => (2.23.140.1.2.3)  (Compliant with Baseline Requirements – Individual identity asserted)

PR #158, opened 2016-07-10
Zzma/ev and ov cert parsing

zzma:
@dadrian @zakird @shoegaze 

EV and OV parsing


ztools/x509/extensions.go
```diff
@@ -128,6 +128,298 @@ type ExtendedKeyUsage []ExtKeyUsage
 
 type CertificatePolicies []asn1.ObjectIdentifier
 
+// The string functions for CertValidationLevel are auto-generated via
+// `go generate <full_path_to_x509_package>` or running `go generate` in the package directory
+//go:generate stringer -type=CertValidationLevel -output=generated_certvalidationlevel_string.go
+type CertValidationLevel int
```
zzma: any issues with this name?

ztools/x509/extensions.go
```diff
@@ -128,6 +128,298 @@ type ExtendedKeyUsage []ExtKeyUsage
 
 type CertificatePolicies []asn1.ObjectIdentifier
 
+// The string functions for CertValidationLevel are auto-generated via
+// `go generate <full_path_to_x509_package>` or running `go generate` in the package directory
+//go:generate stringer -type=CertValidationLevel -output=generated_certvalidationlevel_string.go
```
zzma: this needs to be run manually whenever this `CertValidationLevel` is changed

ztools/x509/extensions.go
shoegaze: Would "classification" be more clear? I'm not really sure.

zzma: I like level because it suggests a linear, tiered classification, rather than a generic classification label. @zakird @dadrian thoughts?

zzma:
bump - changed the slice to a map (set)

zgrab_schema.py
```diff
@@ -130,6 +130,7 @@
     "fingerprint_sha256":Binary(),
     "spki_subject_fingerprint":Binary(),
     "tbs_fingerprint":Binary(),
+    "cert_validation_level": String(),
```
zakird: We probably don't need the "cert_" given that we're in a certificate.

ztools/x509/extensions.go
zakird: Can we just add this to the build process somehow? //cc @dadrian

ztools/x509/extensions.go
dadrian: I think validation levels are what CAs refer to them as.

zakird: I'm alright with validation_level

shoegaze: Suits me

zgrab_schema.py
zzma: 👍

ztools/x509/extensions.go
dadrian: Apparently not a standard thing to do. From the docs:

> Also, if the containing package is intended for import by go get, once the file is generated (and tested!) it must be checked into the source code repository to be available to clients.

ztools/x509/extensions.go
zakird: @zzma we should add this to the zgrab and ztag schema.

ztools/x509/x509.go
```diff
@@ -345,6 +345,19 @@ func getPublicKeyAlgorithmFromOID(oid asn1.ObjectIdentifier) PublicKeyAlgorithm
 	return UnknownPublicKeyAlgorithm
 }
 
+func getMaxCertValidationLevel(oids []asn1.ObjectIdentifier) CertValidationLevel {
+	maxOID := DV
+	for _, oid := range oids {
+		if _, ok := ExtendedValidationOIDs[oid.String()]; ok {
```
dadrian: Shouldn't we also confirm the issuer is correct, since the OID is issuer specific?

dadrian: /cc @zakird

zzma: Technically, yes. But since our sources (Chromium, Mozilla) identify/fingerprint certificates in different ways and CA/Browser (https://cabforum.org/object-registry/) doesn't provide matching certificate fingerprints at all, it's not clear if we would be able to do this accurately, and will probably end up with false negatives. 

On the other hand, if we don't confirm the issuer, we can end up with false positives, but from what I've seen, an issuer has no incentive to use a different CAs proprietary EV or OV OID. 

Side note - this might be a good cert linter task

ztools/x509/extensions.go
zakird: 👎

ztools/x509/x509.go
zakird: 👍

zakird: /cc @josh2059 this should be added to zlint

zakird:
👍  from me. @dadrian: any issues? if not, let's merge.

ztools/x509/x509.go
dadrian: We should file an issue to note that we don't check issuer.

zakird:
Also for reference, this close #147

ztools/x509/extensions.go
dadrian: 👱 🔪 ❤️ 💀

ztools/x509/x509.go
zzma: Created: https://github.com/zmap/zgrab/issues/159

PR #162, opened 2016-07-12
Zzma/set http host header when domain present

zzma:
@zakird @dadrian 

When doing HTTP grabs where an IP, domain are provided, create a connection to <IP>, but set the HTTP host as <domain>. As part of this, I added a functional test that we can use to test other existing/future zgrab logic.

### DIFF
--- a/zlib/grabber.go
+++ b/zlib/grabber.go
@@ -29,6 +29,7 @@ import (
 	"github.com/zmap/zgrab/ztools/scada/siemens"
 	"github.com/zmap/zgrab/ztools/telnet"
 	"github.com/zmap/zgrab/ztools/util"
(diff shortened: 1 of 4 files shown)


zlib/grabber.go
```diff
@@ -223,37 +224,46 @@ func makeHTTPGrabber(config *Config, grabData GrabData) func(string, string) err
...
 		} else {
-			fullURL = "http://" + host + endpoint
+			fullURL = "http://" + urlHost + endpoint
+		}
+
+		var resp *http.Response
+		switch config.HTTP.Method {
+		case "GET":
+			resp, err = client.GetWithHost(fullURL, httpHost)
+		case "HEAD":
```
zzma: re-added support for HEAD requests

PR #164, opened 2016-07-16
Add a couple of getters for Tapdance Client

SergeyFrolov:


dadrian:
Thanks for your contribution and for using ZGrab within TapDance! However, I'd rather not have a public function that returns a private type, since the privateType can't be referred to directly outside of the ztls package. Instead, could we add methods for accessing the sequence number and cipher directly, bypassing the need for exposing `halfConn`?

SergeyFrolov:
Sure, If that matters.
Accessing seq and cipher directly now.

dadrian:
Thanks!

PR #165, opened 2016-07-16
X509 1.7b

SergeyFrolov:
Tried to pull the whole 1.7 beta 2.
Was not sure what I was supposed to pull from upstream, so may have pulled something I shouldn't have. Or may have missed something, notably functions in x509.go, which might be useful otherwise.
Only func SystemCertPool() is used in Tapdance Client.


zakird:
Thanks for the PR. Unfortunately, it looks like this PR conflicts with master in a fair number of places and cannot be merged ( we can't even run tests to see if they would pass). Can you please fix these conflicts and submit a new PR that does not conflict?

zakird:
I'm going to close this PR given that it cannot be merged. However, please don't take this as any indication that we don't want to see 1.7 merged in---we very much do---but I think it makes most sense to submit a new PR when that is a state that we can consider and test.

PR #166, opened 2016-07-21
Pulling root-cert related functions from 1.7

SergeyFrolov:


SergeyFrolov:
4,908 lines out of 5,275 are darwin root certs, so, the PR isn't that big.

PR #168, opened 2016-07-27
Added fields to Certificate to expose UniqueIdentifier fields from the underlying certificate

mhyder13:
Added the fields IssuerUniqueId and SubjectUniqueId as asn1.BitString values to x509.Certificate.
This allows code using this library to detect and access the optional UniqueIdentifier fields from the underlying certificate (if it has them).


zakird:
👍

PR #170, opened 2016-08-03
Added support for reading additional Subject Alternative Name fields

mhyder13:
Support was added for SAN types otherName, directoryName, ediPartyName, uniformResourceIdentifier, and registeredID in certificate parsing. New structs to describe the OtherName and EDIPartyName types were added to the pkix package. Due to limitations of how encoding/asn1 parses DirectoryStrings, ediPartyName and directoryName only support parsing of names encoded with PrintableString and UTF8String. Note that this is the same limitation that applies to parsing the issuer and subject fields of the certificate.

Unit test cases were added to ztools/x509 to test parseSANExtension. Further testing on the ParseCertificate function was done, but not included.

EDIPartyName was verified only with an ASN.1 compiler and multiple checks of the standards since I couldn't find a program that supports it or an example or a properly encoded one. All other types are verified as correctly working with the Windows certificate viewer.


ztools/x509/pkix/pkix.go
```diff
@@ -208,3 +208,17 @@ type RevokedCertificate struct {
 	RevocationTime time.Time
 	Extensions     []Extension `asn1:"optional"`
 }
+
+// OtherName represents the ASN.1 structure of the same name. See RFC
+// 5280, section 4.2.1.6.
+type OtherName struct{
```
zzma: should this have asn1 tags?

ztools/x509/x509.go
```diff
@@ -914,7 +919,12 @@ func parsePublicKey(algo PublicKeyAlgorithm, keyData *publicKeyInfo) (interface{
 	}
 }
 
-func parseSANExtension(value []byte) (dnsNames, emailAddresses []string, ipAddresses []net.IP, err error) {
+/* type OtherName struct{//temp, remove before pull request
```
zzma: 😮

ztools/x509/pkix/pkix.go
mhyder13: The ASN.1 tags of whatever the value is are in the Bytes field of the Value. The Value itself Will always parse out a Tag: 0, Class:2, IsCompound: true since the filed is actually an explicit [0] tag that wraps whatever the OtherName actually contains. I can't have it parse out the contents in detail because it can contain arbitrary data. There was no support for specific OtherName parsing added.

for reference:
OtherName ::= SEQUENCE {
        type-id    OBJECT IDENTIFIER,
        value      [0] EXPLICIT ANY DEFINED BY type-id }

ztools/x509/x509.go
```diff
@@ -948,10 +958,38 @@ func parseSANExtension(value []byte) (dnsNames, emailAddresses []string, ipAddre
 			return
 		}
 		switch v.Tag {
+		case 0:
+			var oName pkix.OtherName
+			_, err = asn1.UnmarshalWithParams(v.FullBytes, &oName, "tag:0")
+			if err == nil{
```
zzma: nit: this is conventionally written as `if err != nil { return } // rest of your code`. 

You can also add the assignment to the if statement in golang:
```
if _, err = asn1.UnmarshalWithParams(v.FullBytes, &oName, "tag:0"); err != nil {
     return
}
```

ztools/x509/pkix/pkix.go
zzma: Gotcha - so there is no need to assign `asn1:"tag:1,explicit"` here for `Value` because it is parsed that way by default?

mhyder13: I did try adding a asn1:"tag:1,explicit", it appears it has no effect when parsing into an asn1.RawValue. The Value field here _is_ the tag, with the contents stored in the Bytes field.

zakird:
For the most part, this looks good to me. @zzma did you have other concerns about the code?

My primary concern is that we don't expose these right now when we output certificates. I didn't originally think this would be problematic---who cares if there's some additional internal parsing. However, now that I thought about this more, I think that was the wrong call. It's going to be very confusing when a certificate fails on a field that isn't even output. It doesn't really allow anyone to understand what's going on and as a user, I would personally assume that this is a bug in the program. Given this, I think we should expose this data all the way through including in the JSON output of the certificate before we merge this PR.

mhyder13:
Do you want me to work on the JSON in this request? If so, how do you want that done? I did try to write this with the idea that this should produce useful data for eventual export, so I certainly agree that exposing this information should be beneficial to users of zgrab etc.

zakird:
I think that makes the most sense. I believe you just need to set this correctly on the out struct.

mhyder13:
I've been looking into how your JSON output currently works, and I noticed a couple of things: It only prints dnsNames, even though rfc822Names and ipAddresses have always been supported. Also, it tries to parse out URLs from the subject pkix.Name, and so it _does_ parse out IP addresses (being valid URLs) that are in the subject and prints those in the JSON. This seems wrong, since all names (according to CA/B) should be in subject alternative name and it is unlikely that the IP address is included as the common name.

Given this information, how should we approach this? I would assume that IPs should go in the current names slice for consistency, but if you have other downstream programs using the names for something other than just display it could cause problems if we start putting other name types in that list. I see three main options: keep the other types separated in their own lists, have everything in the names list, or separate out all the new name types into a second non-web address names list. What would work best for downstream/your preference?

shoegaze:
To me it sounds like your third option will be best for downstream users, as it would be more intuitive to users and less likely to cause hard-to-debug things.

zakird:
So re the first part, people put all sorts of things in the CN and the majority of certificates do not follow CA/B forum guidelines---so it's generally not good to assume that they do. I know with 100% certainty that I've seen certificates with IP in the subject. 

Given the second part, we should output all cases for both CN and SAN. I'll defer to @shoegaze on how we want to handle this in practice.

Re:
>  I see three main options: keep the other types separated in their own lists, have everything in the names list, or separate out all the new name types into a second non-web address names list. What would work best for downstream/your preference?

I'm flexible. Given that this is all becoming non-typed JSON strings, I'm currently thinking let's have them all as one list and just output them accordingly. But we might come to regret that too. 
@shoegaze @dadrian @zzma What do you think?

mhyder13:
So there are certainly certificates with IPs in the subject, but I think it is a bug to assume that an IP address _will_ be in the subject and ignore the ones in SAN.

zakird:
Agree---we should handle all cases in both fields regardless of what's in the other one.

zakird:
It might help if we had some concrete data to look at. @mhyder13 can you sketch out what you would imagien the JSON output would be for the three cases?

zakird:
OK after thinking about this more and looking at the current situation, I have a slightly different take.

For SAN: split out into individual types.
For CN: always just output a string

In order words, let's keep doing what we're already doing and just add the new types.

dadrian:
So SAN, unlike CN, actually has structured data: dnsNames, ipAddresses, etc, defined in the ASN.1 for the SAN extension. Therefore, I think it's pretty clear that we should just add the new/missing fields as subfields of `subject_alternative_name` in the JSON. The `common_name` field we pretty much solely interpret as a string, and I think we should leave it that way, since it's _much_ less structured than the SAN.

The `names` field exists to make it easier to find all the names in the certificate in one spot, since not all certificates put all names in the SAN. We run CN through govalidator to see if there is any URL that might have been put inside the CN, since URL parsing is hard and the CN isn't as strongly typed.

I believe we have a clear path forward:

1. Add the new SAN types as subfields of the existing `subject_alternative_name` JSON object.
2. Continue to use govalidator to  find any extra URLs that are in the CN
3. Take the deduplicated union of the URLs from the CN and the URL-like fields from the SAN, and place that in `names`. Run all these through govalidator to enforce they're actually a URL and not a random string, or a raw encoding of a JPG shoved into the otherName field of a CN, etc.

For an example certificate, that might give an output of something like this:

```json
"parsed": {
    "subject": {
        "common_name": "davidadrian.org"
     },
    "subject_alternative_name": {
        "dns_names": [ 
            "www.davidadrian.org",
            "herp derp not a real URL"
        ],
        "ip_addresses": [
            "1.2.3.4"
        ],
        "other_names": [
            "base64encodedstring=="
        ]
    },
    "names": [
        "davidadrian.org",
        "www.davidadrian.org",
        "1.2.3.4"
    ]
}
```

Does this make sense to other people?

zakird:
I agree with @dadrian. So shall it be.

mhyder13:
Example output in the new format:
```
"subject_alt_name":{
    "directory_names":[{"organization":["Extreme Discord"]}],
    "dns_names":["*.gov.us","gov.us"],
    "edi_party_names":[{"name_assigner":"Mother Nature","party_name":"partyName"}],
    "email_addresses":["admin@gov.us"],
    "ip_addresses":["11.10.10.11"],
    "other_names":[{"id":"1.3.6.1.4.1.11483.46","value":"AgMIKgk="}],
    "registered_ids":["1.3.6.1.4.1.11483.442"],
    "uniform_resource_identifiers":["https://gov.us/home"]
}
```
You can also see in the build output that duplicates are now eliminated from the "Names" field. 
Before:
```"names":["www.jhalderm.com", "jhalderm.com@proxy.dreamhost.com", "www.jhalderm.com","jhalderm.com", "pad.jhalderm.com", "*.pad.jhalderm.com"]```
After:
```"names":["www.jhalderm.com", "jhalderm.com@proxy.dreamhost.com", "jhalderm.com", "pad.jhalderm.com", "*.pad.jhalderm.com"]```
(spaces added to make it wrap properly)

**Comment by mhyder13** [comment]
I believe that I have incorporated everything we spoke about above, so the code should be ready to be looked over again. Let me know if you see anything I missed.

**Comment by zakird** [comment]
@shoegaze matt, would you be willing to look over this?

File: ztools/x509/extensions.go [hunk 4]
```diff
@@ -63,9 +63,42 @@ type BasicConstraints struct {
...
+	EDIPartyNames  []pkix.EDIPartyName      `json:"edi_party_names,omitempty"`
+	EmailAddresses []string                 `json:"email_addresses,omitempty"`
+	IPAddresses    []net.IP                 `json:"ip_addresses,omitempty"`
+	OtherNames     []pkix.OtherName         `json:"other_names,omitempty"`
+	RegisteredIDs  []string                 `json:"registered_ids,omitempty"`
+	URIs           []string                 `json:"uniform_resource_identifiers,omitempty"`
+}
+
+func (san SubjectAltName) MarshalJSON() ([]byte, error) {
+	jsan := jsonSubjectAltName{
```
shoegaze: > `jsan`

I see what you did there.

ztools/x509/json.go
```diff
@@ -260,6 +260,21 @@ func (c *Certificate) MarshalJSON() ([]byte, error) {
 			jc.Names = append(jc.Names, name)
 		}
 	}
+	
+	for _, name := range c.URIs {
```
shoegaze: Question for @zakird and @dadrian: Do we need to handle urls with *'s in them? If so this code won't work.

zakird: Is the `*` in the FQDN or the path? I don't think they're valid in the FQDN, but probably are in the path?

shoegaze: I think if they're in the path it will be fine, it's just if they're in the FQDN that it breaks govalidator. It appears that they can't appear in FQDNs so it should be fine.

shoegaze:
Okay, everything looks alright to me, I'm a bit tired, so it might be wise to have another pair of eyes on it, but as far as I'm concerned we're good to go.

zakird:
TODO: we also need to add this to zgrab and ztag schema before we merge this.

ztools/x509/extensions.go
dadrian: This should have a pointer receiver.

`func (san *SubjectAltName) MarshalJSON()`

dadrian: I would also love you forever if you implemented `func (san *SubjectAltName) UnmarshalJSON(b []byte) error`, but that's not necessary.

ztools/x509/json.go
dadrian: `*` should only appear in DNS names, not URI's, right?

shoegaze: Right, I just wanted a sanity check.

ztools/x509/json.go
```diff
@@ -322,3 +337,18 @@ func (c *Certificate) MarshalJSON() ([]byte, error) {
 
 	return json.Marshal(jc)
 }
+
+func purgeNameDuplicates(names []string) (out []string){
```
dadrian: You can just use an `map[string]struct{}` for this (e.g. a map to nothing), and then iterating over it will give you a deduplicated list.

mhyder13: That's roughly how this already works, except that it is using a map[string]bool instead.

ztools/x509/extensions.go
mhyder13: That change will be included in the next commit. I'm working on UnmarshalJSON(). It's almost done, but pkix.Name is also missing that function, so I need to write one for that before I can push it out.

zakird:
@zzma is taking ownership of fixing this up and taking care of zgrab schema

mhyder13:
The build error is coming from a compiler version mismatch: you are still running 1.5.2 and there is a reference to a constant added in 1.6.  I'm not sure how you want to resolve that, but it could be fixed by just using the integer literal 2. There is no error on go versions >= 1.6.

Additionaly, pkix.Name still needs an UnmarshalJSON() to finish out the changes.

zakird:
Let's define the constant ourselves.

zzma:
@zakir - anything holding us back from upping the version of golang in Travis to 1.7.0?

zakird:
I'm fine adding the newer versions. I'm not sure if we want to pull others
though. At the very least we should have 1.6 too.

mhyder13:
@zzma The current unmarshal version doesn't seem to be entirely working:
(from top to bottom: struct, json output, unmarshal result, struct == result)

> {[{[US] [Mother Nature] [Everything] [Tallahassee] [FL] [3210 Holly Mill Run] [30062] []  Earth [] []}] [www.name1.com net.name2.net] [] [webmaster@www.name1.com the@net.name2.net] [] [] [1.8.5.27] [http://www.name1.com urn:isbn:5589678]}
> 
> {"directory_names":[{"common_name":["Earth"],"country":["US"],"locality":["Tallahassee"],"organization":["Mother Nature"],"organizational_unit":["Everything"],"postal_code":["30062"],"province":["FL"],"street_address":["3210 Holly Mill Run"]}],"dns_names":["www.name1.com","net.name2.net"],"email_addresses":["webmaster@www.name1.com","the@net.name2.net"],"registered_ids":["1.8.5.27"],"uniform_resource_identifiers":["http://www.name1.com","urn:isbn:5589678"]}
> 
> {[{[US] [Mother Nature] [] [Tallahassee] [FL] [] [] []   [{2.5.4.6 US} {2.5.4.10 Mother Nature} {2.5.4.7 Tallahassee} {2.5.4.8 FL}] []}] [www.name1.com net.name2.net] [] [webmaster@www.name1.com the@net.name2.net] [] [] [1.8.5.27] [http://www.name1.com urn:isbn:5589678]}
>  false

We appear to be losing OrganizationalUnit, PostalCode, StreetAddress, and CommonName in the process, at minimum. We are probably losing it somewhere in this line: `if err := json.Unmarshal(b, &jName); err != nil {` since jsonName has a special marshaler and there is no matching unmarshal function to undo the special stuff it does.

dadrian:
I'm a little late to the party here, and it looks like the decision was made off chat, but I'm all for bumping up to 1.6 or higher. The scan workers are both easy to update and already running 1.6.2.

zzma:
@mhyder13 @dadrian This should be good to go now. Both `Marshal` and `Unmarshal` working for `pkix.Name`. Some sample output: 
```
  "subject_alt_name": {
    "directory_names": [
      {
        "common_name": ["davidadrian.org"],
        "country": ["US"],
        "locality": ["Ann Arbor"],
        "organization": [
          "University of Michigan",
          "Computer Science Department"
        ],
        "province": ["MI"],
        "serial_number": ["12345678910"]
      }
    ],
    "dns_names": [
      "*.gov.us",
      "gov.us"
    ],
    "edi_party_names": [
      {
        "name_assigner": "Mother Nature",
        "party_name": "partyName"
      }
    ],
    "email_addresses": ["admin@gov.us"],
    "ip_addresses": ["11.10.10.11"],
    "other_names": [
      {
        "id": "1.3.6.1.4.1.11483.46",
        "value": "AgMIKgk="
      }
    ],
    "registered_ids": ["1.3.6.1.4.1.11483.442"],
    "uniform_resource_identifiers": ["https:\/\/gov.us\/home"]
  }
```

zakird:
This LGTM offhand minus missing schema in ZGrab and ZTag. @zzma I'm fine with you merging this when you're comfortable, but would like to look at the ZTag schema before it's pushed.

dadrian:
@zzma why are the forward slashes escaped in the above example?

zzma:
@dadrian - sorry that is an oversight - the output format was copy-pasted from a test logger output string that was doing the escaping. The actual output will not be slash-escaped

zgrab_schema.py
```diff
@@ -21,6 +21,19 @@
...
+    "serial_number":ListOf(String()),
+    "country":ListOf(String()),
+    "locality":ListOf(String()),
+    "province":ListOf(String()),
+    "street_address":ListOf(String()),
+    "organization":ListOf(String()),
+    "organizational_unit":ListOf(String()),
+    "postal_code":ListOf(String()),
+    "domain_component":ListOf(String()),
+})
```
zakird: Is this different than what's in Subject and Issuer/

zzma: @zakird - nope! will clean up

zzma:
This is ready to merge. I added schema changes and more tests. The travis failure is from decreased coverage, but all the tests pass.

dadrian:
Also, do we have example output for `names`?

zzma:
@dadrian `names` is not output in JSON. it is just a giant list of every field in the `pkix.Name` in AttributeTypeAndValue form

shoegaze:
@zzma I think it's output in the pkix json.

dadrian:
I mean the `names` field that gets added to the X509 object that has every FQDN across SAN + CN. Or is that not in this diff?

zzma:
@shoegaze - I'm pretty sure `pkix.Name` does not output `Names`

zzma:
@dadrian - here is a sample: 

```
"names":["test@santest.com","dns1.test.com","dns2.test.com","http://watchit.com/","1.2.3.4"]
```

shoegaze:
Oops, my bad. I was looking at it earlier and confused myself.

zzma:
Any other thoughts? If not, I'm going to go ahead and merge. There are some people who are blocked on this PR currently.

mhyder13:
Are email addresses actually getting through into the names output as in that example? Is that supposed to be doing that?

dadrian:
IMHO we should only put DNSNames and IP addresses in `names`

zakird:
I could go either way. The field is somewhat made up anyway and I could see it going either way.

dadrian:
I'm revising my opinion to just be DNSNames + DNSNames in CN, because that's the common case for what people search for, and the part that browsers use.

zzma:
yeah - I think that we keep `names` as it is right now, but we can make other changes in subsequent PRs

zakird:
SGTM

PR #172, opened 2016-08-26
UnmarshalJSON() fix for pkix.OtherName

mhyder13:
This change causes Value.FullBytes to be correctly filled in when an OtherName is unmarshaled.

You might want to squash commit this one, there are some test commits and the merge from zmap in the history.


ztools/x509/pkix/json.go
```diff
@@ -200,6 +200,10 @@ func (o *OtherName) UnmarshalJSON(b []byte) error {
 		IsCompound: true,
 		Bytes:      oName.Value,
 	}
+	o.Value.FullBytes, err = asn1.Marshal(o.Value)
+	if err != nil{
```
zakird: This looks like there's a missing space.

ztools/x509/pkix/json.go
```diff
@@ -175,10 +175,10 @@ func (o *OtherName) MarshalJSON() ([]byte, error) {
 	return json.Marshal(oName)
 }
 
-func (o *OtherName) UnmarshalJSON(b []byte) error {
+func (o *OtherName) UnmarshalJSON(b []byte) (err error) {
```
zakird: I don't understand why this was changed.

ztools/x509/pkix/json.go
zzma: Also - this `if err != nil .... return nil` can be replaced by a simple `return`

ztools/x509/pkix/json.go
```diff
@@ -175,10 +175,10 @@ func (o *OtherName) MarshalJSON() ([]byte, error) {
...
 	var oName jsonOtherName
 
-	if err := json.Unmarshal(b, &oName); err != nil {
+	if err = json.Unmarshal(b, &oName); err != nil {
 		return err
```
zzma: when you're naming return variables, I think it's conventional to simply write `return` which makes it more clear that there are named return variables, rather than `return err` which somewhat defeats the point of a named return variable.

zakird:
LGTM modulo @zzma's comments, which should be addressed before we merge this.

ztools/x509/pkix/json.go
```diff
@@ -270,7 +270,6 @@ func (n *Name) UnmarshalJSON(b []byte) error {
 	n.Names = appendATV(n.Names, jName.PostalCode, oidPostalCode)
 	n.Names = appendATV(n.Names, jName.DomainComponent, oidDomainComponent)
 
-
```
mhyder13: This line courtesy of gofmt.

ztools/x509/pkix/json.go
```diff
@@ -200,6 +201,10 @@ func (o *OtherName) UnmarshalJSON(b []byte) error {
 		IsCompound: true,
 		Bytes:      oName.Value,
 	}
+	o.Value.FullBytes, err = asn1.Marshal(o.Value)
+	if err != nil {
+		return
+	}
```
zakird: This return logic can be simplified.

ztools/x509/pkix/json.go
```diff
@@ -175,20 +175,21 @@ func (o *OtherName) MarshalJSON() ([]byte, error) {
...
+	if err = json.Unmarshal(b, &oName); err != nil {
+		return
 	}
 
 	arcs := strings.Split(oName.Id, ".")
 	oid := make(asn1.ObjectIdentifier, len(arcs))
 
 	for i, s := range arcs {
-		tmp, err := strconv.ParseInt(s, 10, 32)
+		var tmp int64
```
zakird: This variable does not need to be defined. Use := in the next line.

mhyder13: It was changed because it won't compile otherwise. I could change it to `tmp, e := strconv.ParseInt(s, 10, 32)` if you like. That should work.

zakird: Alright, let's leave it.

ztools/x509/pkix/json.go
zakird: This still needs to be resolved. Once it has been, lgtm to merge.

mhyder13:
@zakird The return logic at the end has been fixed and I pulled in the upstream changes to allow a merge.

PR #174, opened 2016-08-27
Pull Request: adding more name constraints support, and some bug fixes.

zhengping12:
x509.go: 
added name constraints parsing for EdiPartyName, x400Address, DirectoryName, and RegisteredIds
wrapped the name constraint field along with a maximum and minimum in structs
extensions.go:
added more support to marshalJson and unmarshalJson of name contraints.
fixed two bugs in the previous existing marshalJson code
verify.go:
modified name constraint types to match the change in x509.go
extensions_test.go:
added condition to skip hidden files from test certs list, to avoid the program reading .DS_Store file on macs.
added name constraint marshalJson and unmarshalJson test
json_test.go:
added condition to skip hidden files from test certs list, to avoid the program reading .DS_Store file on macs.


zakird:
Hi @zhengping12,
It looks like this branch currently conflicts with the master branch and cannot be merged in. Can you please merge master into your branch? Second, there are some formatting oddities in the PR. Can you please run `go fmt` on your branch?

zhengping12:
@zakird Done.

ztools/x509/json_test.go
```diff
@@ -33,6 +33,9 @@ func (s *JSONSuite) SetUpTest(c *C) {
 	}
 
 	for _, test := range tests {
+		if test.Name()[0] == 46 {
```
dadrian: What's the significance of 46?

dadrian:
I'll take a closer look tomorrow, but looks reasonable overall.

zhengping12:
@dadrian 46 is ascii of "." , which tells it to skip hidden files. In particular .DS_Store that is literally in every folder on MacOS

ztools/x509/json_test.go
zhengping12: @dadrian 46 is ascii of "." , which tells it to skip hidden files. In particular .DS_Store that is literally in every folder on MacOS

zakird:
NOTE: This must have schema changes before it can be merged.

ztools/x509/json_test.go
dadrian: What files are you looking for? I'd rather whitelist on suffix than blacklist on some random subset of prefixes.

zhengping12: ok, fixed.

ztools/x509/extensions.go
```diff
@@ -140,44 +140,153 @@ func (san *SubjectAltName) UnmarshalJSON(b []byte) error {
...
+	}
+	for _, email := range ncJson.PermittedEmailDomains {
+		nc.PermittedEmailDomains = append(nc.PermittedEmailDomains, GeneralSubtreeString{Data: email})
+	}
+	for _, ipNet := range ncJson.PermittedIPAddresses {
+		ip, ipnet, err := net.ParseCIDR(ipNet)
+		if err != nil {
+			return err
+		}
+		ipnet.IP = ip
```
dadrian: Why do we set `ipnet.IP` here?

ztools/x509/extensions.go
```diff
@@ -140,44 +140,153 @@ func (san *SubjectAltName) UnmarshalJSON(b []byte) error {
...
+	}
+	for _, email := range ncJson.ExcludedEmailDomains {
+		nc.ExcludedEmailDomains = append(nc.ExcludedEmailDomains, GeneralSubtreeString{Data: email})
+	}
+	for _, ipNet := range ncJson.ExcludedIPAddresses {
+		ip, ipnet, err := net.ParseCIDR(ipNet)
+		if err != nil {
+			return err
+		}
+		ipnet.IP = ip
```
dadrian: Same question.

ztools/x509/extensions_test.go
```diff
@@ -33,6 +33,9 @@ func (s *ExtensionsSuite) SetUpTest(c *C) {
 	}
 
 	for _, test := range tests {
+		if test.Name()[len(test.Name())-5:] != ".cert" {
```
dadrian: Just use https://golang.org/pkg/strings/#HasSuffix :)

ztools/x509/extensions_test.go
```diff
@@ -48,7 +51,7 @@ func (s *ExtensionsSuite) SetUpTest(c *C) {
 
 func (s *ExtensionsSuite) TestEncodeDecodeSAN(c *C) {
 	for _, cert := range s.parsedCerts {
-		if (cert.Issuer.CommonName != "SAN Test"){
+		if cert.Issuer.CommonName != "SAN Test" {
```
dadrian: Are you sure this won't cause false positives, e.g. tests passing that shouldn't?

ztools/x509/json.go
```diff
@@ -339,16 +339,16 @@ func (c *Certificate) MarshalJSON() ([]byte, error) {
 	return json.Marshal(jc)
 }
 
-func purgeNameDuplicates(names []string) (out []string){
+func purgeNameDuplicates(names []string) (out []string) {
 	hashset := make(map[string]bool, len(names))
-	for _, name := range names{
-		if _, inc := hashset[name]; !inc{
+	for _, name := range names {
+		if _, inc := hashset[name]; !inc {
```
dadrian: I'd still prefer empty struct for this, but w/e.

ztools/x509/json_test.go
```diff
@@ -33,6 +33,9 @@ func (s *JSONSuite) SetUpTest(c *C) {
 	}
 
 	for _, test := range tests {
+		if test.Name()[len(test.Name())-5:] != ".cert" {
```
dadrian: Same comment about `HasSuffix`

ztools/x509/extensions.go
zhengping12: The ipnet returned from ParseCIDR is truncated of the unnecessary least significant bits.
While ip contains the full IP. I'm not sure if this was needed.

ztools/x509/testdata/name.constraint.test.cert
```diff
@@ -0,0 +1,32 @@
+-----BEGIN CERTIFICATE-----
```
dadrian: Where did this certificate come from? IIRC, you can add random text data to the bottom of the cert without the parse functionality failing, it's probably worth describing what this certificate is, or at least appending the output `openssl x509 -text -noout` for this certificate.

ztools/x509/x509.go
```diff
@@ -589,17 +589,29 @@ type Certificate struct {
 	IPAddresses    []net.IP
 	RegisteredIDs  []asn1.ObjectIdentifier
 
+	// Certificate Policies values
+	QualifierId                [][]asn1.ObjectIdentifier
+	ExplicitTexts              [][]asn1.RawValue
+	NoticeRefOrgnization       [][]asn1.RawValue
+	NoticeRefNumbers					 [][]NoticeNumber
```
dadrian: What's up with this spacing?

ztools/x509/verify_test.go
```diff
@@ -5,411 +5,411 @@
...
-    return chainStr
+	var chainStr string
+	for _, cert := range chain {
+		if len(chainStr) > 0 {
+			chainStr += " -> "
+		}
+		chainStr += nameToKey(&cert.Subject)
+	}
+	return chainStr
 }
```
dadrian: Why are there all these formatting changes? Did you run gofmt? Was this not previously gofmt? Is your editor changing spaces to tabs?

dadrian: Err rather, tabs to spaces.

ztools/x509/x509.go
```diff
@@ -625,6 +637,44 @@ type Certificate struct {
 	SignedCertificateTimestampList []*ct.SignedCertificateTimestamp
 }
 
+type NoticeNumber []int
+
+type GeneralSubtreeString struct {
+	Data string
+	Max  int
```
dadrian: What are the significance of `Max` and `Min`?

ztools/x509/x509.go
```diff
@@ -821,7 +886,9 @@ type nameConstraints struct {
 }
 
 type generalSubtree struct {
-	Value asn1.RawValue `asn1:"optional"`
+	Value   asn1.RawValue `asn1:"optional"`
+	Minimum int           `asn1:"tag:0,default:0,optional"`
```
dadrian: Probably want to standardize `Min`/`Max` vs `Minimum`/`Maximum`

ztools/x509/extensions_test.go
zhengping12: Zane wrote this. This is just a unit test. He created a cert to go with this unit test that has a CommonName "SAN Test"

ztools/x509/json.go
zhengping12: I didn't touch this file, besides the gofmt change.

ztools/x509/extensions_test.go
zhengping12: Thanks, I'll fix this

ztools/x509/x509.go
```diff
@@ -1135,51 +1202,87 @@ func parseCertificate(in *certificate) (*Certificate, error) {
 				for _, subtree := range constraints.Permitted {
 					switch subtree.Value.Tag {
```
dadrian: Are these tags ASN.1 constants defined for Name Constraints?

ztools/x509/x509.go
```diff
@@ -1135,51 +1202,87 @@ func parseCertificate(in *certificate) (*Certificate, error) {
...
+					case 5:
+						var ediName pkix.EDIPartyName
+						_, err = asn1.Unmarshal(subtree.Value.Bytes, &ediName)
+						if err != nil {
+							return out, err
+						}
+						out.PermittedEdiPartyNames = append(out.PermittedEdiPartyNames, GeneralSubtreeEdi{Data: ediName, Max: subtree.Maximum, Min: subtree.Minimum})
 					case 7:
 						switch len(subtree.Value.Bytes) {
 						case net.IPv4len * 2:
```
dadrian: I'm not sure I follow why this is `len * 2`, but it looks like that was already the case, so \shrug.

ztools/x509/x509.go
```diff
@@ -1274,8 +1377,29 @@ func parseCertificate(in *certificate) (*Certificate, error) {
...
+				out.QualifierId = make([][]asn1.ObjectIdentifier, len(policies))
+				out.ExplicitTexts = make([][]asn1.RawValue, len(policies))
+				out.NoticeRefOrgnization = make([][]asn1.RawValue, len(policies))
+				out.NoticeRefNumbers = make([][]NoticeNumber, len(policies))
 				for i, policy := range policies {
 					out.PolicyIdentifiers[i] = policy.Policy
+					// parse optional Qualifier for zlint
+					for _, qualifier := range policy.Qualifiers {
+						out.QualifierId[i] = append(out.QualifierId[i], qualifier.PolicyQualifierId)
+						if qualifier.PolicyQualifierId.Equal(asn1.ObjectIdentifier{1, 3, 6, 1, 5, 5, 7, 2, 2}) {
```
dadrian: Is it possible to make this `asn1.ObjectIdentifier` a constant, or at least a private named var?

dadrian:
Alright, I've gone through and added comments. Overall looks pretty good, sorry it took me so long to get to it. In the future, if you're ever blocking on me, feel free to ping me on email / chat.

I'm not familiar with the details of NameConstraint encoding, so I'm assuming you're handing that correctly. I'd like to see an example output for certificates with none/most/all of the new fields you're adding.

ztools/x509/verify_test.go
zhengping12: Yea, I just ran gofmt, it wasn't previously formatted.

ztools/x509/x509.go
zhengping12: It is the same for all generalNames, so it also applies to SAN and other extensions that use generalName.

ztools/x509/testdata/name.constraint.test.cert
zhengping12: Should I append the output to the end of the certificate file?

ztools/x509/extensions.go
dadrian: What do you mean by truncated? Can you give an example?

ztools/x509/extensions_test.go
dadrian: If you only have one certificate for each test, why not just explicitly load that certificate in the test?

ztools/x509/json.go
dadrian: Ah, np then.

ztools/x509/testdata/name.constraint.test.cert
dadrian: Try the top or the bottom, whatever works. Example: https://cs.chromium.org/chromium/src/net/data/ssl/certificates/self-signed-invalid-sig.pem?q=self-signe&sq=package:chromium&l=1

ztools/x509/verify_test.go
mhyder13: I noticed that code written wholly on Zhengping's computer usually seems to have 2 spaces per indent instead of tabs but match the format if it was already using tabs. You you might want to adjust your settings/always gofmt files before commit.

ztools/x509/x509.go
mhyder13: I don't think you can make it a constant because it is actually a []int, but the other way should work depending on where the variable is.

ztools/x509/x509.go
mhyder13: From RFC5280:
>    "Within this profile, the minimum and maximum fields are not used with
>    any name forms, thus, the minimum MUST be zero, and maximum MUST be
>    absent.  However, if an application encounters a critical name
>    constraints extension that specifies other values for minimum or
>    maximum for a name form that appears in a subsequent certificate, the
>    application MUST either process these fields or reject the
>    certificate."


From X.509:

> The minimum field specifies the upper bound of the area within the subtree. All names whose final name component is above the level specified are not contained within the area. A value of minimum equal to zero (the default) corresponds to the base, i.e., the top node of the subtree. For example, if minimum is set to one, then the subtree excludes the base node but includes subordinate nodes.
> 
> The maximum field specifies the lower bound of the area within the subtree. All names whose last component is below the level specified are not contained within the area. A value of maximum of zero corresponds to the base, i.e., the top of the subtree. An absent maximum component indicates that no lower limit should be imposed on the area within the subtree. For example, if maximum is set to one, then the subtree excludes all nodes except the subtree base and its immediate subordinates. 

Essentially, they have something to do with depth in the subtree, but they "aren't used" in the internet pkix.

ztools/x509/verify_test.go
dadrian: Indeed. Please do set up your text editor to run goimports / gofmt on save. There's instructions for pretty much every major editor trivially available online.

zhengping12: Ok, i'll get gofmt-before-save. But changes to verify_test.go was solely due to gofmt if I remember correctly.

ztools/x509/extensions.go
zhengping12: if we call
ip, ipNet, err := net.ParseCIDR("192.168.1.1/24")
we get 
ip ==  IP {192, 168, 1, 1}
ipNet == {IP:IP{192, 168, 1, 0}, Mask:IPMask{255, 255, 255, 0}}

ztools/x509/extensions_test.go
zhengping12: in the SetUpTest functions, all certificates are parsed in one go and stored into ExtensionsSuite struct. If we were to load the certificate in the test, it will get parsed a second time.

zhengping12:
Thanks for the input @dadrian . I went and fixed the issues you mentioned.

ztools/x509/extensions_test.go
dadrian: Yes, but unless I'm mistaken (I may very well be), there's not a single test that needs _all_ of them.

ztools/x509/extensions.go
dadrian: Ah, I see. We're slightly abusing the semantics of the object then, but I think that's OK because the semantics will only be abused when the certificate has an invalid CIDR on it, anyway. (192.168.1.1/24 isn't a valid CIDR).

tl;dr I'm OK with this.

dadrian:
So it'd be nice if we could only load the certificates in the tests they're actually used, but I don't think that's a blocking issue. If you are motivated to go fix that, make a new PR :).

LGTM.

dadrian:
Actually, one last nit: can you post a couple of different example outputs with various NameConstraint fields set?

ztools/x509/extensions_test.go
zhengping12: I'm not sure either, I only added the name constraints test. If we were to do the parsing inside of the test, we need to restructure the testing code. But again, I didn't come up them.

zhengping12:
-----------Almost empty name constraints------------
X509v3 Name Constraints:
                Excluded:
                  DNS:banned.com
------------------------------------------------------

-----------Half filled name constraints---------------
X509v3 Name Constraints:
                Permitted:
                  Registered ID1.2.3.4
                  email:email
                  email:LulMail
                  DirName: C = US, O = UIUC, OU = ECE, L = Champaign, ST = IL, street = 601 Wright St, postalCode = 61820, CN = uiuc.net
                Excluded:
                  DNS:banned.com
                  IP:192.168.1.1/255.255.0.0
------------------------------------------------------

-----------Fully filled name constraints---------------
 X509v3 Name Constraints:
                Permitted:
                  email:good_email@gg.com
                  email:LulMail
                  DNS:permitted.com
                  DirName: C = US, O = UIUC, OU = ECE, L = Champaign, ST = IL, street = 601 Wright St, postalCode = 61820, CN = uiuc.net
                  IP:74.125.224.72/255.255.0.0
                Excluded:
                  email:bad_email@gg.com
                  email:LulMail
                  DNS:banned.com
                  DirName: C = US, O = Umich, OU = CS, L = Ann Arbor, ST = MI, street = 500 State St, postalCode = 48109, CN = umich.net
                  IP:192.168.1.1/255.255.0.0
------------------------------------------------------

EdiPartyName is left out since openssl command line tool doesn't support this.

dadrian:
Sorry! I missed that you responded here. Do feel free to ping me if that happens.

Also, I realize now that I wasn't clear. I meant, could you post the ZGrab JSON output, not the OpenSSL output.

zhengping12:
@dadrian Don't worry about it. Here they are, partly and fully filled JSON.

{"critical":false,"permitted_email_addresses":["email","LulMail"],"permitted_directory_names":[{"common_name":["uiuc.net"],"country":["US"],"locality":["Champaign"],"organization":["UIUC"],"organizational_unit":["ECE"],"postal_code":["61820"],"province":["IL"],"street_address":["601 Wright St"]}],"permitted_registred_id":["1.2.3.4"],"excluded_names":["banned.com"],"excluded_ip_addresses":["192.168.1.1/16"]}

{"critical":false,"permitted_names":["permitted.com"],"permitted_email_addresses":["good_email@gg.com","LulMail"],"permitted_ip_addresses":["74.125.224.72/16"],"permitted_directory_names":[{"common_name":["uiuc.net"],"country":["US"],"locality":["Champaign"],"organization":["UIUC"],"organizational_unit":["ECE"],"postal_code":["61820"],"province":["IL"],"street_address":["601 Wright St"]}],"excluded_names":["banned.com"],"excluded_email_addresses":["bad_email@gg.com","LulMail"],"excluded_ip_addresses":["192.168.1.1/16"],"excluded_directory_names":[{"common_name":["umich.net"],"country":["US"],"locality":["Ann Arbor"],"organization":["Umich"],"organizational_unit":["CS"],"postal_code":["48109"],"province":["MI"],"street_address":["500 State St"]}]}

zakird:
@dadrian this looks in line with our schema.

dadrian:
Thanks for the PR and dealing with all the fixes!

zhengping12:
Thanks for all the suggestions!  :)

Issue #143, opened 2016-06-25
Add SHA256(TBSCertificate No Poison)


zakird:

dadrian:
No poison? Does this have something to do with name redaction?

zakird:
CT Poison.

shoegaze:
Ref here: http://www.alvestrand.no/objectid/submissions/1.3.6.1.4.1.11129.2.4.3.html

shoegaze:
opened https://github.com/zmap/zgrab/tree/NoPoisonFingerprint to solve this, will submit pr shortly.

shoegaze:
Fixed by #179

PR #179, opened 2016-09-03
Added a tbs fingerprint without poison extension 

shoegaze:
for percert/cert linking to address #143.


shoegaze:
So this works, I think it's ready to go.

Issue #175, opened 2016-08-28
Ports not handled correctly during redirects


zakird:
There's a lot of this going on in the logs currently:

`
Aug 28 00:17:41.435 [ERROR] banner-grab: Could not connect to remote host http://14.85.213.90:80/: Get http://14.85.213.90:80:8899: dial tcp: too many colons in address 14.85.213.90:80:8899
Aug 28 00:17:41.442 [ERROR] banner-grab: Could not connect to remote host http://125.130.39.251:80/: Get http://125.130.39.251:80:8899: dial tcp: too many colons in address 125.130.39.251:80:8899
Aug 28 00:17:41.455 [ERROR] banner-grab: Could not connect to remote host http://14.78.150.138:80/: Get http://14.78.150.138:80:8899: dial tcp: too many colons in address 14.78.150.138:80:8899
`

We should check if we're handling ports correctly during redirects.

PR #182, opened 2016-09-10
set host without port

zzma:
Fixes: https://github.com/zmap/zgrab/issues/175

@zakird @dadrian 

When testing, the error was caused because servers were returning `Location` headers that had extra ports attached to them, i.e. `ip.add.re.ss:80:8899`. The underlying reason for this is that when we have an IP address HTTP zgrab, we set the `Host` HTTP header to include port number, which we shouldn't.


zzma:
Will fix failed tests tonight

ztools/http/request.go
```diff
@@ -449,7 +450,11 @@ func NewRequestWithHost(method, urlStr, host string, body io.Reader) (*Request,
 	}
 
 	if host == "" {
-		host = u.Host
+		if hostWithoutPort, _, err := net.SplitHostPort(u.Host); err != nil {
```
zakird: So what happens here if there isn't a port number? Can that ever happen?

zzma: It just returns the host if there is no port number

zakird: Oh perfect! I missed that when reading the description.

zakird:
I'm fine with merging this once tests are passing.

zzma:
@zakird - looking at this a bit more, it seems like setting the port is actually a thing we should be doing generally, according to the RFC (https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html), and that the errors we get are due to poorly implemented servers.

One possibility for what we do is only include the port if it is a non-default port for the protocol (i.e. 80 for http and 443 for https). Thoughts?

zakird:
That seems fine to me.

zakird:
This doesn't seem to actually work. I ran

`echo "209.118.99.170,google.com" | ./zgrab --port=80 --http --http-max-redirects=10` and got

`{"ip":"209.118.99.170","domain":"google.com","timestamp":"2016-09-12T11:46:31-05:00","data":{"http":{}},"error":"missing port in address google.com"}`

zzma:
Yeah - still working on some tests to handle the domain case.

zzma:
I'll bump when ready

zakird:
Ah OK, I didn't realize that, I just only tests were remaining. Sorry for
the noise.

zzma:
no problemo. Bumping now. This should properly handle the cases of IP, IP+domain, domain. FYI - with this change the domains that were yielding multiple-port redirects now redirect to a single port, but many of the redirect connections time out (which is also what we see in the browser).

zlib/grabber.go
```diff
@@ -193,6 +195,16 @@ func makeTLSConfig(config *Config, urlHost string) *ztls.Config {
...
 
+func usingDefaultPort(scheme string, port uint16) bool {
+	return (scheme == "https" && port == 443) || (scheme == "http" && port == 80)
+}
+
+// Given a string of the form "host", "host:port", or "[ipv6::address]:port",
+// return true if the string includes a port, does not validate port
+func containsPort(host string) bool {
+	return strings.LastIndex(host, ":") > strings.LastIndex(host, "]")
+}
```
dadrian: https://golang.org/pkg/net/#SplitHostPort

zzma: Saw that, but decided that it would be better to just implement this rather than do `SplitHostPort` and compare on a custom/specific error string that may change in the future.

dadrian: Wouldn't _any_ error string be enough?

zzma: I want to distinguish between the case where `SplitHostPort` fails on an absent port, in which case we keep calm and carry on, and where `SplitHostPort` fails with some other hostname parsing error, in which case we want to abort and return the error.

dadrian:
LGTM

zakird:
Something still seems very awry. I just ran
`echo "209.118.99.170,google.com" | ./zgrab --port=80 --http --http-max-redirects=10 | jq .`

which should fetch Google's normal page, but this fails with the following error: `Sep 13 12:41:13.095 [ERROR] banner-grab: Could not connect to remote host http://209.118.99.170:80--http-max-redirects=10: Get http://www.google.com/: stopped after 0 redirects`

And it's output is even stranger:

```
{
  "ip": "209.118.99.170",
  "domain": "google.com",
  "timestamp": "2016-09-13T12:41:12-05:00",
  "data": {
    "http": {
      "redirect_response_chain": [
        {
          "status_line": "301 Moved Permanently",
          "status_code": 301,
          "protocol": {
            "name": "HTTP/1.1",
            "major": 1,
            "minor": 1
          },
          "headers": {
            "cache_control": [
              "public, max-age=2592000"
            ],
            "content_length": [
              "219"
            ],
            "content_type": [
              "text/html; charset=UTF-8"
            ],
            "expires": [
              "Thu, 13 Oct 2016 17:41:14 GMT"
            ],
            "location": [
              "http://www.google.com/"
            ],
            "server": [
              "gws"
            ],
            "unknown": [
              {
                "key": "date",
                "value": [
                  "Tue, 13 Sep 2016 17:41:14 GMT"
                ]
              }
            ],
            "x_frame_options": [
              "SAMEORIGIN"
            ],
            "x_xss_protection": [
              "1; mode=block"
            ]
          },
          "body": "<HTML><HEAD><meta http-equiv=\"content-type\" content=\"text/html;charset=utf-8\">\n<TITLE>301 Moved</TITLE></HEAD><BODY>\n<H1>301 Moved</H1>\nThe document has moved\n<A HREF=\"http://www.google.com/\">here</A>.\r\n</BODY></HTML>\r\n",
          "body_sha256": "IXju3Vcjpqwi6U7Fm9zZkinIfzYjdT9eGZZ4JC8OkN4=",
          "content_length": 219,
          "request": {
            "url": {
              "scheme": "http",
              "host": "209.118.99.170:80--http-max-redirects=10"
            },
            "method": "GET",
            "headers": {
              "user_agent": [
                "Mozilla/5.0 zgrab/0.x"
              ]
            },
            "host": "google.com"
          }
        }
      ]
    }
  },
  "error": "Get http://www.google.com/: stopped after 0 redirects"
}
{
  "port": 80,
  "success_count": 0,
  "failure_count": 1,
  "total": 1,
  "start_time": "2016-09-13T12:41:12-05:00",
  "end_time": "2016-09-13T12:41:13-05:00",
  "duration": 0,
  "senders": 1000,
  "timeout": 10,
  "tls_version": null,
  "mail_type": null,
  "ca_file_name": null,
  "sni_support": true
}
```

For example, note the value of `host`.

zzma:
@zakird - you need to set the --http endpoint, there is no default value currently. The command should look like: `echo "209.118.99.170,google.com" | ./zgrab --port=80 --http="/" --http-max-redirects=10 | jq .` which seems to return the right thing.

zakird:
golang arg parsing--

zakird:
LGTM

Issue #186, opened 2016-09-24
certificates with TLDs trusted don't correctly get added to names field


zakird:
Example: https://censys.io/certificates/719c282a51e935051e88bf6115dda0731da21c0e12c08e6bcea36078e83e4966

PR #188, opened 2016-09-29
Cert perspectives fix

### LINKED ISSUE #171: names field doesn't capture redacted names
Because ?.domain.com isn't technically a valid name, this won't end up being put into the `names` array. We should make our domain validator less strict to accept these. The names field is already a custom construct and given that we want to support easily searching for certificates, it makes the most sense to just include these.

umbernhard:
Fixed #171 and #186

### DIFF
--- a/ztools/x509/json.go
+++ b/ztools/x509/json.go
@@ -229,29 +229,23 @@ func (c *Certificate) MarshalJSON() ([]byte, error) {
 	jc.SubjectKeyInfo.KeyAlgorithm = c.PublicKeyAlgorithm
 
 	// Include all subject names, DNS names there are
-	for _, obj := range c.Subject.Names {
+	name := c.Subject.CommonName
+	isValid := false
 
-		switch name := obj.Value.(type) {
-		case string:
+	if len(name) > 2 && name[0] == '*' {
+		isValid = govalidator.IsURL(name[2:])
+	} else if len(name) > 2 && name[0] == '?' {
+		isValid = isValid || govalidator.IsURL(name[2:])
+	} else if !strings.Contains(name, ".") {
+		// If this is just a TLD cert, it's valid
+		isValid = true
+	} else {
+		isValid = govalidator.IsURL(name)
+	}
 
-			isValid := false
-
-			if len(name) > 2 && name[0] == '*' {
-				isValid = govalidator.IsURL(name[2:])
-			} else if len(name) > 2 && name[0] == '?' {
-				isValid = isValid || govalidator.IsURL(name[2:])
-			} else if !strings.Contains(name, ".") {
-				// If this is just a TLD cert, it's valid
-				isValid = true
-			} else {
-				isValid = govalidator.IsURL(name)
-			}
-
-			// Check that this is actually a url and not something else
-			if isValid {
-				jc.Names = append(jc.Names, name)
-			}
-		}
+	// Check that this is actually a url and not something else
+	if isValid {
+		jc.Names = append(jc.Names, name)
 	}
 
 	for _, name := range c.DNSNames {
--- a/ztools/x509/json.go
+++ b/ztools/x509/json.go
@@ -11,6 +11,7 @@ import (
 	"encoding/asn1"
 	"encoding/json"
 
+	"strings"
 	"time"
 
 	"github.com/asaskevich/govalidator"
@@ -233,16 +234,27 @@ func (c *Certificate) MarshalJSON() ([]byte, error) {
 		switch name := obj.Value.(type) {
 		case string:
 
-			flag := false
+			isValid := false
 
 			if len(name) > 2 && name[0] == '*' {
-				flag = govalidator.IsURL(name[2:])
+				isValid = govalidator.IsURL(name[2:])
 			} else {
-				flag = govalidator.IsURL(name)
+				isValid = govalidator.IsURL(name)
+			}
+
+			// If this is just a TLD cert, it's valid
+			if !strings.Contains(name, ".") {
+				isValid = true
+			}
+
+			if name[0] == '?' {
+				isValid = isValid && govalidator.IsURL(name[2:])
+			} else {
+				isValid = isValid && govalidator.IsURL(name)
 			}
 
 			// Check that this is actually a url and not something else
-			if flag {
+			if isValid {
 				jc.Names = append(jc.Names, name)
 			}
 		}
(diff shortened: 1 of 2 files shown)


umbernhard:
Tested against https://censys.io/certificates/b354890665fd2e05f13a6a029eeb701b7f8dd6f45f14d68f7bfa2f9a7d41470c

and 

https://censys.io/certificates/719c282a51e935051e88bf6115dda0731da21c0e12c08e6bcea36078e83e4966

ztools/x509/json.go
```diff
@@ -228,37 +229,41 @@ func (c *Certificate) MarshalJSON() ([]byte, error) {
...
-				flag = govalidator.IsURL(name)
-			}
+	name := c.Subject.CommonName
+	isValid := false
+
+	if len(name) > 2 && name[0] == '*' {
+		isValid = govalidator.IsURL(name[2:])
+	} else if len(name) > 2 && name[0] == '?' {
+		isValid = isValid || govalidator.IsURL(name[2:])
+	} else if !strings.Contains(name, ".") {
```
umbernhard: It won't, because of line 232 in this commit: https://github.com/zmap/zgrab/commit/adccb4b0c2d79ad4d171ecbcc9f7921d2631284f#diff-b69ed0926a8809fc22de87e608774fe9R232

ztools/x509/json.go
```diff
@@ -228,37 +229,41 @@ func (c *Certificate) MarshalJSON() ([]byte, error) {
...
-
-			if len(name) > 2 && name[0] == '*' {
-				flag = govalidator.IsURL(name[2:])
-			} else {
-				flag = govalidator.IsURL(name)
-			}
+	name := c.Subject.CommonName
+	isValid := false
+
+	if len(name) > 2 && name[0] == '*' {
```
dadrian: Shouldn't we look at `name[1]` ever?

umbernhard: so the reason for this is that govalidator won't validate *.blah.com. Will *a.blah.com ever be possible?

zakird: Knowing how _terrible_ people are at making certs..... probably.

umbernhard: Right. So should we consider that valid or not? 

@dadrian's right that I should look at names[1], there could be a weird case of *abc.blah.com, where this code would check c.blah.com as a result, which is *technically* valid but meaningless. I'll make this more robust.

ztools/x509/json.go
umbernhard: Never mind, I'm dumb. Will fix.

ztools/x509/json.go
```diff
@@ -228,37 +229,40 @@ func (c *Certificate) MarshalJSON() ([]byte, error) {
...
+	isValid := false
+
+	// Check for wildcards and redacts, ignore malformed urls
+	if len(name) > 2 && name[0] == '*' && name[1] == '.' {
+		isValid = govalidator.IsURL(name[2:])
+	} else if len(name) > 2 && name[0] == '?' && name[1] == '.' {
+		isValid = isValid || govalidator.IsURL(name[2:])
+	} else {
+		isValid = govalidator.IsURL(name)
+	}
```
dadrian: Can we extract this whole chunk of logic to a function, since it's repeated in the `DNSNames` loop later.

ztools/x509/json.go
```diff
@@ -228,37 +229,40 @@ func (c *Certificate) MarshalJSON() ([]byte, error) {
...
+			// If this is just a TLD cert, it's valid
+			isValid = true
 		} else {
-			flag = govalidator.IsURL(name)
+			isValid = govalidator.IsURL(name)
 		}
 
-		if flag {
+		if isValid {
 			jc.Names = append(jc.Names, name)
```
dadrian: This doesn't look deduplicated.

ztools/x509/json.go
```diff
@@ -228,37 +229,40 @@ func (c *Certificate) MarshalJSON() ([]byte, error) {
...
-			} else {
-				flag = govalidator.IsURL(name)
-			}
+	name := c.Subject.CommonName
+	isValid := false
+
+	// Check for wildcards and redacts, ignore malformed urls
+	if len(name) > 2 && name[0] == '*' && name[1] == '.' {
+		isValid = govalidator.IsURL(name[2:])
+	} else if len(name) > 2 && name[0] == '?' && name[1] == '.' {
```
umbernhard: This still leaves us open to *abc.x.y, since it would leave us with bc.x.y, which govalidator would see as valid. Is there like a regex or something to let us specify StartsWith(name, "*?[.]" or something?

ztools/x509/json.go
umbernhard: good point, I can do that.

ztools/x509/json.go
```diff
@@ -228,37 +229,40 @@ func (c *Certificate) MarshalJSON() ([]byte, error) {
...
-		flag := false
+		isValid := false
 
-		if len(name) > 2 && name[0] == '*' {
-			flag = govalidator.IsURL(name[2:])
+		// Check wildcards, redacts, malformed, and tlds
+		if len(name) > 2 && name[0] == '*' && name[1] == '.' {
+			isValid = govalidator.IsURL(name[2:])
+		} else if len(name) > 2 && name[0] == '?' && name[1] == '.' {
+			isValid = isValid || govalidator.IsURL(name[2:])
```
umbernhard: Realized that this `or` is no longer needed

ztools/x509/json.go
umbernhard: IIRC, deduplication happens later.

ztools/x509/json.go
umbernhard: Hmm, actually I'm realizing the logic isn't repeated exactly. For DNSNames we also check for TLDs

umbernhard: Figured out a way to make it work.

ztools/x509/json.go
umbernhard: Ah yes, see https://github.com/zmap/zgrab/blob/certPerspectivesFix/ztools/x509/json.go#L283

umbernhard:
@dadrian I think I made the fix you requested.

ztools/x509/json.go
dadrian: Whomp, I meant:
```
if strings.StartsWith(name, "*.") {
    isValid = govalidator.IsURL(name[2:])
} else if strings.StartsWith(name, "?.") {
    isValid = govalidator.IsURL(name[2:])
} else {
    isValid = govalidator.IsURL(name)
}
```

umbernhard: Is done.

ztools/x509/json.go
```diff
@@ -356,3 +334,15 @@ func purgeNameDuplicates(names []string) (out []string) {
 	}
 	return
 }
+
+func isValid(name string) (ret bool) {
+
```
dadrian: Maye name `isValidName`?

dadrian:
Is `?.?.example.com` allowed? @zakird

zakird:
I mean there aren't any real rules.

dadrian:
Should we handle that?

dadrian:
I'm good to merge if @zakird is.

umbernhard:
I would suggest checking for the ?.? case and at least noting that it happens, but I don't really know how big a deal it is.

umbernhard:
or if it would materially change anything

zakird:
This looks good to me. I'm not too worried about the ?.? situation, I'm not seeing evidence of this happening in practice.

PR #191, opened 2016-10-06
revert the small change that broke 1.5 support

zzma:


zzma:
oops. Sorry didn't mean to PR this.

PR #193, opened 2016-10-10
Adding Issuer DN checking

bvandersloot:


ztools/x509/x509.go
```diff
@@ -778,6 +778,10 @@ func (c *Certificate) CheckSignatureFrom(parent *Certificate) (err error) {
 
 	// TODO(agl): don't ignore the path length constraint.
 
+	if parent.Subject.String() != c.Issuer.String() {
+		return errors.New("Mis-match issuer/subject")
+	}
+
```
zakird: This probably doesn't belong in this function. Your signature can be valid even with a issuer-subject mismatch.

zakird: Actually I take this back. While this function is named *terribly*, there _are_ other checks in it already. :(

ztools/zct/x509/x509.go
```diff
@@ -619,6 +619,10 @@ func (c *Certificate) CheckSignatureFrom(parent *Certificate) (err error) {
 
 	// TODO(agl): don't ignore the path length constraint.
 
+	if parent.Subject.String() != c.Issuer.String() {
+		return errors.New(fmt.Sprintf("Mis-match issuer/subject (%s!=%s)", parent.Subject.String(), c.Issuer.String()))
+	}
+
```
zakird: We can just check SubjectDN and IssuerDN instead of doing this twice.

bvandersloot: Those are not parsed out in this version of x509.

PR #197, opened 2016-11-08
Added support to source the grabber traffic from a specific IP address / interface on the machine.

timapril:
The IP in question is checked to make sure that it is configured
for the device to reduce the ability to spoof.

### DIFF
--- a/main.go
+++ b/main.go
@@ -1,5 +1,7 @@
 /*
- * ZGrab Copyright 2015 Regents of the University of Michigan
+ * ZGrab
+ *   Copyright 2015 Regents of the University of Michigan
+ *   Copyright 2016 Akamai Technologies Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not
  * use this file except in compliance with the License. You may obtain a copy
@@ -20,6 +22,7 @@ import (
 	"flag"
 	"io"
 	"io/ioutil"
+	"net"
 	"os"
 	"runtime"
 	"strings"
@@ -45,6 +48,7 @@ var (
 	timeout                       uint
 	tlsVersion                    string
 	rootCAFileName                string
+	sourceIP                      string
 )
 
 // Module configurations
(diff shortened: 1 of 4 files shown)


main.go
```diff
@@ -140,6 +145,36 @@ func init() {
 		zlog.Fatalf("--connections-per-host must be in the range [0,50]")
 	}
 
+	// If a source IP is specified, verify that the host is advertising the IP
+	// and has a chance to receive the responses.
+	if sourceIP != "" {
+		config.LocalAddressSet = false
+		providedAddr := net.ParseIP(sourceIP)
```
dadrian: On error, this can return nil. We should `Fatalf` if that is the case.

main.go
```diff
@@ -140,6 +145,36 @@ func init() {
...
+		for idx, i := range localInterfaces {
+			addresses, adddressesErr := i.Addrs()
+			if adddressesErr != nil {
+				zlog.Fatalf("Error getting Local addresses for interface %d: %s", idx, adddressesErr.Error())
+			}
+			for _, address := range addresses {
+				castAddress := address.(*net.IPNet)
+				if castAddress.Contains(providedAddr) {
+					tcpAddr := &net.TCPAddr{
+						IP: address.(*net.IPNet).IP,
```
dadrian: This can be `IP: castAddress.IP`

main.go
timapril: Good catch, I just noticed that my errcheck addon was turned off. I will update that shortly.

main.go
timapril: This appears to work. I thought I tried that and got an odd error. I will roll that into the new version.

timapril:
The machine that I have been testing on has 2 interfaces, one that has 2 v4 addresses and a v6 address and then another interface that is not enabled. With this change, I am able to send all of the traffic out of one of the virtual interfaces on one NIC.

zakird:
I'm hesitant to split copyright for this project as proposed. We have yet so far managed to avoid this and longterm, I'd prefer to not set a precedent where the copyright for small chunks of code are held by different groups. Does Akamai require this, or can copyright be assigned to Michigan?

dadrian:
I'm closing this PR until the copyright issue is resolved.

timapril:
We have submitted a CLA so the copyright issue should be addressed now. @zakird will probably have more information.

PR #201, opened 2016-11-16
Defer resp.Body.Close() in HTTP

dadrian:
I believe we were leaking memory on errored connections.


zzma:
LGTM - thanks for tracking that down!

zakird:
LGTM. Let's run it and see what happens.

PR #206, opened 2016-12-17
Parse ecdsa curve

zakird:


zgrab_schema.py
```diff
@@ -56,7 +56,6 @@
         "fingerprint_sha256":Binary(),
         "key_algorithm":SubRecord({
             "name":String(doc="Name of public key type, e.g., RSA or ECDSA. More information is available the named SubRecord (e.g., rsa_public_key)."),
-            "oid":String(doc="OID of the public key on the certificate. This is helpful when an unknown type is present. This field is reserved and not current populated.")
```
dadrian: Can we safely remove this? Aren't we sending it up currently?

ztools/x509/json.go
```diff
@@ -280,12 +287,13 @@ func (c *Certificate) MarshalJSON() ([]byte, error) {
 		keyMap["gy"] = params.Gy.Bytes()
 		keyMap["x"] = key.X.Bytes()
 		keyMap["y"] = key.Y.Bytes()
+		keyMap["curve"] = key.Curve.Params().Name
```
dadrian: Oh snap! This did not exist back in 2014.

ztools/x509/json.go
```diff
@@ -339,10 +349,9 @@ func isValidName(name string) (ret bool) {
 
 	// Check for wildcards and redacts, ignore malformed urls
 	if strings.HasPrefix(name, "?.") || strings.HasPrefix(name, "*.") {
-		ret = govalidator.IsURL(name[2:])
+		ret = isValidName(name[2:])
```
dadrian: Not really sure why we need to make this recursive.

zakird:
Because otherwise you don't allow ?.?.x.y which commonly appears on
redacted certificates

zakird:
Re OID : no. It was never populated.

Issue #210, opened 2017-01-17
--tls should default to port 443 or require a port argument


paul-pearce:
Currently `zgrab --tls` defaults to port 80, which seems worse than either requiring a port or defaulting to port 443.

umbernhard:
Turns out this is tricky. In order to default to 443 for TLS but still allow people to force port 80, we need to roll our own golang flag (see [here](https://stackoverflow.com/questions/35809252/check-if-flag-was-provided-in-go)).

Then we have to figure out how to get a default value into that. Or we could just assume that if people want to use TLS they're never going to force port 80. If we do that, it's really easy.

zakird:
ZGrab2, which is superseding this project has default ports for modules. Closing.

PR #211, opened 2017-01-17
Added code to address #210.

umbernhard:
* implemented a flag for the port
* change default dynamically based on whether or not we're using TLS

### DIFF
--- a/main.go
+++ b/main.go
@@ -284,6 +284,8 @@ func init() {
 	// default to port 443 if we are using TLS
 	if !portFlag.set && config.TLS {
 		portFlag.Set("443")
+	} else if !portFlag.set {
+		portFlag.Set("80")
 	}
 	config.Port = portFlag.value
 
--- a/main.go
+++ b/main.go
@@ -22,6 +22,7 @@ import (
 	"io/ioutil"
 	"os"
 	"runtime"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -255,10 +277,17 @@ func init() {
 	config.Encoding = encoding
 
 	// Validate port
-	if portFlag > 65535 {
-		zlog.Fatal("Port", portFlag, "out of range")
+	if portFlag.value > 65535 {
```
paul-pearce: It's impossible to ever hit this now, as the cast has already occured.

main.go
```diff
@@ -32,6 +33,25 @@ import (
...
 
+type port struct {
+	value uint16
+	set   bool
+}
+
+func (pf *port) Set(x string) error {
+	value, err := strconv.Atoi(x)
+	if err != nil {
+		zlog.Fatal("Port", x, "is invalid")
```
paul-pearce: Missing spaces in the error msg

main.go
```diff
@@ -32,6 +33,25 @@ import (
 	"github.com/zmap/zgrab/ztools/ztls"
 )
 
+type port struct {
+	value uint16
+	set   bool
```
paul-pearce: It seems heavy handed to keep a bool around for the lifetime of the port just to identify if it is set. You could accomplish this with a sentential value, a ptr, or separate data structure that only lives the lifetime of the configuration.

paul-pearce:
I'm not a maintainer on this so obviously defer to @dadrian or @zakird, but my gut reaction would be to revert back to uint for the port, and use 0 as a sentinel value (treat `port ==0` as `port.set == false` in this code). It's not perfect, as if a user passes in port 0 it will behave as though they passed in nothing. But if the user passes in 0 they are going to have bigger problems on their hands.

main.go
umbernhard: I thought I didn't need them? If I do we should fix line 281 too.

dadrian:
Why don't we just require `--port` be passed in? This doesn't seem particularly worth fixing right now, without doing a much larger overhaul of how ZGrab does modules / flags in general.

zakird:
I agree with @dadrian here.

PR #220, opened 2017-02-01
Pull in ALPN tls extension from golang tls

SergeyFrolov:
Pulling in this extension: https://tools.ietf.org/html/rfc7301
Next Protocol Negotiation, which is already present in ztls, is predecessor of Application Layer Protocol Negotiation.
NPN: client says it supports NPN, but doesn't specify the supported protocols in Client Hello; server responds with list of protocols, from which client can choose.
ALPN: client suggests supported protocols in Client Hello.


ztools/ztls/handshake_client.go
```diff
@@ -643,9 +644,29 @@ func (hs *clientHandshakeState) processServerHello() (bool, error) {
...
+
+	if !clientDidALPN && serverHasALPN {
+		c.sendAlert(alertHandshakeFailure)
+		return false, errors.New("tls: server advertised unrequested ALPN extension")
+	}
+
+	if serverHasNPN && serverHasALPN {
+		c.sendAlert(alertHandshakeFailure)
+		return false, errors.New("tls: server advertised both NPN and ALPN extensions")
+	}
```
dadrian: Is this for sure an error?

ztools/ztls/handshake_client.go
```diff
@@ -643,9 +644,29 @@ func (hs *clientHandshakeState) processServerHello() (bool, error) {
...
+
+	if serverHasALPN {
```
dadrian: Should we also be checking `clientHasALPN` here?

ztools/ztls/handshake_client.go
SergeyFrolov: I just copied code from https://golang.org/src/crypto/tls/handshake_client.go
I haven't found definite answer in rfc, however this piece of code was written by Adam Langley https://go.googlesource.com/go/+blame/master/src/crypto/tls/handshake_client.go, who was involved in drafting both NPN and ALPN

ztools/ztls/handshake_client.go
SergeyFrolov: We checked if `!clientDidALPN && serverHasALPN { ` case before

dadrian:
@zakird Any concerns?

ztools/ztls/handshake_messages.go
```diff
@@ -576,6 +631,13 @@ func (m *serverHelloMsg) marshal() []byte {
 		extensionsLength += 1
 		numExtensions++
 	}
+	if alpnLen := len(m.alpnProtocol); alpnLen > 0 {
+		if alpnLen >= 256 {
+			panic("invalid ALPN protocol")
```
zakird: Should this result in a panic, or just a failed handshake?

ztools/ztls/handshake_messages.go
```diff
@@ -99,6 +101,17 @@ func (m *clientHelloMsg) marshal() []byte {
 		extensionsLength += 1
 		numExtensions++
 	}
+	if len(m.alpnProtocols) > 0 {
+		extensionsLength += 2
+		for _, s := range m.alpnProtocols {
+			if l := len(s); l == 0 || l > 255 {
+				panic("invalid ALPN protocol")
```
zakird: Should this be a panic or just a failed handshake?

ztools/ztls/common.go
```diff
@@ -78,6 +78,7 @@ const (
 	extensionSupportedCurves      uint16 = 10
 	extensionSupportedPoints      uint16 = 11
 	extensionSignatureAlgorithms  uint16 = 13
+	extensionALPN                 uint16 = 16
```
SergeyFrolov: extensionApplicationLayerProtocolNegotiation is a very long string

ztools/ztls/handshake_client.go
```diff
@@ -643,9 +644,29 @@ func (hs *clientHandshakeState) processServerHello() (bool, error) {
...
+	if !clientDidNPN && serverHasNPN {
 		c.sendAlert(alertHandshakeFailure)
-		return false, errors.New("server advertised unrequested NPN extension")
+		return false, errors.New("tls: server advertised unrequested NPN extension")
+	}
+
+	if !clientDidALPN && serverHasALPN {
+		c.sendAlert(alertHandshakeFailure)
+		return false, errors.New("tls: server advertised unrequested ALPN extension")
+	}
```
SergeyFrolov: I just copied (all) code from https://golang.org/src/crypto/tls/handshake_client.go
I haven't found definite answer in rfc, however this piece of code was written by Adam Langley https://go.googlesource.com/go/+blame/master/src/crypto/tls/handshake_client.go, who was involved in drafting both NPN and ALPN

dadrian:
@SergeyFrolov Can you respond to @zakird's remaining questions?

ztools/ztls/handshake_server.go
```diff
@@ -167,13 +167,20 @@ Curves:
...
+		if selectedProto, fallback := mutualProtocol(hs.clientHello.alpnProtocols, c.config.NextProtos); !fallback {
+			hs.hello.alpnProtocol = selectedProto
+			c.clientProtocol = selectedProto
+		}
+	} else {
+		// Although sending an empty NPN extension is reasonable, Firefox has
+		// had a bug around this. Best to send nothing at all if
+		// config.NextProtos is empty. See
+		// https://code.google.com/p/go/issues/detail?id=5445.
+		if hs.clientHello.nextProtoNeg && len(config.NextProtos) > 0 {
```
SergeyFrolov: Perhaps. I am not sure if it is worthwhile to change, though: now ztls has a benefit of being identical to main tls' code. Do you have a strong opinion on this?

zakird: If it's the original code, I guess just leave it.

ztools/ztls/handshake_messages.go
SergeyFrolov: I think it is appropriate to panic() here. This panic is not triggered by server, but by client, when he's doing something very wrong: you'd want to know about misconfiguration.

ztools/ztls/handshake_messages.go
SergeyFrolov: I think it is appropriate to panic() here. This panic is not triggered by server, but by client, when he's doing something very wrong: you'd want to know about misconfiguration.

ztools/ztls/handshake_client.go
```diff
@@ -54,6 +54,7 @@ func (c *Conn) clientHandshake() error {
 		supportedPoints:      []uint8{pointFormatUncompressed},
 		nextProtoNeg:         len(c.config.NextProtos) > 0,
 		secureRenegotiation:  true,
+		alpnProtocols:        c.config.NextProtos,
```
SergeyFrolov: I believe this was done for backward compatibility.
Before ALPN(e.g. NPN): Client says they support NPN, server responds with list of protocols, client chooses the first one from c.config.NextProtos
With ALPN: Client says they support NPN and sends c.config.NextProtos to server
So, interface hasn't changed for user

ztools/ztls/handshake_messages.go
SergeyFrolov: If you think panic() is not appropriate, then it would be easy to remove that check entirely, and I can modify the PR for it.
That will result in client sending invalid ALPN to server, which might or might not fail the handshake.

PR #221, opened 2017-02-02
Fix SNI insertion

mtury:
With inputs of the form `<IP>,<domain_name>`, and no `lookup-domain` option, the SNI extension in the ClientHellos sent by `zgrab` contained `<IP>:<port>` instead of the domain name. This is non-standard and probably unwanted behavior.

### DIFF
--- a/zlib/grabber.go
+++ b/zlib/grabber.go
@@ -214,7 +214,7 @@ func makeHTTPGrabber(config *Config, grabData *GrabData) func(string, string, st
 
 		var tlsConfig *ztls.Config
 		if config.TLS {
-			tlsConfig = makeTLSConfig(config, urlHost)
+			tlsConfig = makeTLSConfig(config, httpHost)
 		}
 
 		transport := &http.Transport{


dadrian:
@zzma Can you look at this?

zzma:
@mtury - I am unable to replicate the issue doing HTTPS on port 443. We  Can you show me the command you are using? I have been testing with this command: `echo "216.58.192.206,google.com" | zgrab --port 443 --tls`

dadrian:
@zzma you confirmed correct behavior with Wireshark?

paul-pearce:
I dug into this a little (it's relevant to my USENIX push). The erroneous behavior only occurs when `config.NoSNI` is set. This can be set in a few different ways.

The following invocation produces SNI field with the port included (it probably shouldnt use SNI at all):

`echo "216.239.38.21" | ./zgrab -tls -port=443 --http="/"`

image

Oddly enough, the following invocation produces the bogus SNI as well (but fails to connect):
`echo "216.239.38.21,censys.io" | ./zgrab -tls -port=443 --http="/"`

But this has the correct SNI field set and work fine:
`echo "216.239.38.21,censys.io" | ./zgrab -tls -port=443`

With this PR, the following produces a correct SNI:

`echo "216.239.38.21,censys.io" | ./zgrab -tls -port=443 -senders=1000 --http="/"`

`echo "216.239.38.21,censys.io" | ./zgrab -tls -port=443 -senders=1000`

And the following produce no SNI at all:

`echo "216.239.38.21" | ./zgrab -tls -port=443 -senders=1000 --http="/"`

`echo "216.239.38.21" | ./zgrab -tls -port=443 -senders=1000`

So, LGTM.

zzma:
@paul-pearce thanks for the closer look! Looks good to me

PR #222, opened 2017-02-02
Fix renegotiation extension unmarshall

mtury:
When `unmarshal`-ling ClientHellos, the `renegotiation_info` extension failed to be identified.

### DIFF
--- a/ztools/ztls/handshake_messages.go
+++ b/ztools/ztls/handshake_messages.go
@@ -460,7 +460,7 @@ func (m *clientHelloMsg) unmarshal(data []byte) bool {
 				m.signatureAndHashes[i].signature = d[1]
 				d = d[2:]
 			}
-		case extensionRenegotiationInfo + 1:
+		case extensionRenegotiationInfo:
 			if length != 1 || data[0] != 0 {
 				return false
 			}


dadrian:
@mtury I'm confused. What exactly is the problem here, and how does this fix it?

mtury:
Unless I misunderstood the method (I confess I have no idea how this `+ 1` got lost here), when parsing ClientHello extensions with `unmarshal()`, a few lines meant to be related to the `renegotiation_info` extension were executed. But these were triggered by a `0xff02` extension identifier, instead of the rightful `0xff01` which actually defaulted to nothing.

Because the real `renegotiation_info` extensions were not recognized, the sanity check was not performed and the `secureRenegotiation` attribute to the internal ClientHello remained set to `false`. In particular, this would prevent the extension to reappear as expected in the encoding returned by `marshal()`.

dadrian:
Oh, wow. I was reading the PR backwards. I have no idea where that `+ 1` came from. It looks like it's been there we first introduced the TLS code.

Thanks for the fix!

zakird:
Oh wow, I did the exact same thing. Good catch.

PR #223, opened 2017-02-02
Add external ClientHello support

mtury:
This commit enables importing raw ClientHellos to be sent during the `zgrab` scans.

When providing such information with the `raw-client-hello` option, `zgrab` will only rewrite the SNI extension with the current domain name to be queried. The ClientHello contents will stay otherwise untouched (well, except for the extensions order). As a side effect, other options like `chrome-ciphers` or `tls-session-ticket` are ignored.

Main motivations behind this:
- some extensions may not be known to `zgrab`, but we might want to send them nonetheless;
- `zgrab` options do not cover every field/value to be written in the ClientHello probe. I've been using `scapy` for designing TLS messages, which is more exhaustive and flexible IMHO. (The [PR](https://github.com/secdev/scapy/pull/476) has been in review for a few months now.)

With the fix from PR #221, you may test this commit with something like the command below. Wireshark should show that the outgoing ClientHello contains an ALPN extension (unsupported by `zgrab` for now).
`echo "8.8.8.8,www.google.com" | zgrab --tls --port 443 --raw-client-hello dummy_clientHello.txt`

[dummy_clientHello.txt](https://github.com/zmap/zgrab/files/748575/dummy_clientHello.txt)

### DIFF
--- a/main.go
+++ b/main.go
@@ -99,7 +99,7 @@ func init() {
 	flag.BoolVar(&config.S7, "s7", false, "Send some Siemens S7 data")
 	flag.BoolVar(&config.NoSNI, "no-sni", false, "Do not send domain name in TLS handshake regardless of whether known")
 
-	flag.StringVar(&clientHelloFileName, "raw-client-hello", "", "Provide a raw ClientHello to be sent; only the SNI will be rewritten (implies --tls)")
+	flag.StringVar(&clientHelloFileName, "raw-client-hello", "", "Provide a raw ClientHello to be sent; only the SNI will be rewritten")
 
 	flag.BoolVar(&config.ExportsOnly, "export-ciphers", false, "Send only export ciphers")
 	flag.BoolVar(&config.ExportsDHOnly, "export-dhe-ciphers", false, "Send only export DHE ciphers")
@@ -347,7 +347,6 @@ func init() {
 			zlog.Fatal(err)
 		} else {
 			config.ExternalClientHello = clientHello
-			config.ForceClientHello = true
 		}
 	}
 }
--- a/main.go
+++ b/main.go
@@ -47,6 +47,7 @@ var (
 	tlsVersion                    string
 	rootCAFileName                string
 	prometheusAddress             string
+	clientHelloFileName           string
 )
 
 // Module configurations
@@ -97,6 +98,8 @@ func init() {
 	flag.BoolVar(&config.Fox, "fox", false, "Send some Niagara Fox Tunneling data")
 	flag.BoolVar(&config.S7, "s7", false, "Send some Siemens S7 data")
 	flag.BoolVar(&config.NoSNI, "no-sni", false, "Do not send domain name in TLS handshake regardless of whether known")
+
+	flag.StringVar(&clientHelloFileName, "raw-client-hello", "", "Provide a raw ClientHello to be sent; only the SNI will be rewritten (implies --tls)")
 
 	flag.BoolVar(&config.ExportsOnly, "export-ciphers", false, "Send only export ciphers")
 	flag.BoolVar(&config.ExportsDHOnly, "export-dhe-ciphers", false, "Send only export DHE ciphers")
@@ -337,6 +340,16 @@ func init() {
 	}
 	logger := zlog.New(logFile, "banner-grab")
 	config.ErrorLog = logger
+
+	// Open TLS ClientHello, if applicable
+	if clientHelloFileName != "" {
+		if clientHello, err := ioutil.ReadFile(clientHelloFileName); err != nil {
+			zlog.Fatal(err)
+		} else {
+			config.ExternalClientHello = clientHello
+			config.ForceClientHello = true
+		}
+	}
 }
 
 func main() {
--- a/ztools/ztls/common.go
+++ b/ztools/ztls/common.go
@@ -354,8 +354,7 @@ type Config struct {
 	// Explicitly set Client random
 	ClientRandom []byte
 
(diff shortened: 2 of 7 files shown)


zakird:
This seems *very* similar to #224. Do we need both approaches?

dadrian:
I'm a little confused. What is the goal here / can you give a concrete example of what you want to use this for? I like the idea, but I'm not sure this is the best way to go about it.

Since you still unmarshal the external client hello, aren't we going to end up ignoring every extension that is unknown to ZGrab? Off hand, this seems like an obscure way to to avoid a configuration file or some command-line flags? What do we actually gain with this?

I'm inclined to go with an approach more similar to #224.

mtury:
I agree that PR #224 provides virtually the same capability. The main difference lies in the way you want to specify the ClientHello contents to ZGrab. @bvandersloot 's proposition relies on a configuration file while mine uses raw data.

Extensibility was an issue here, as I wanted to be able to send extensions unknown to ZGrab, like ALPN, Padding, or unregistered extensions. One could also send TLS 1.3 extensions before they have been registered in ZGrab. In my PR, these extensions are transparently preserved by the unmarshal operation through the `unknownExtensions` slice added to both `clientHelloMsg` and `serverHelloMsg` structs. Unless I'm mistaken, PR #224 users would need to add to the code every unknown extension to ZGrab in order to achieve the same result.

Flexibility was another goal. Up to now ZGrab provided little control over the ClientHello random, cipher suites, extensions, etc. I don't think using a flag for every field would be practical for the user, or even realistic from an implementer point of view. Using a configuration file as PR #224 is a better way to go, yet I still find it clumsy. The way fields are marshaled sometimes allows limited room for variation, and the config validation may not be immediate unless you're confident about the names and structs defined in ZGrab.

Obviously I'm biased from my usage of Scapy, but I think it's way more comfortable to write a ClientHello message (e.g. for checking server Padding support) in a Python interpreter, with something as simple as `TLSClientHello(ext= TLS_Ext_Padding(padding="\x00"*200) )`. ZGrab is a great tool for scanning, but I wouldn't say the same on the matter of designing TLS packets, and I don't think it should be expected either. This is a different job after all.

Finally, I was concerned about exportability and datasets comparison. While providing a configuration file to explain what was sent to each server offers easily readable content, it still has to be processed by ZGrab before the ClientHello is generated. What if the code has been changed since the time of the measurements? PR #221 and PR #222 somehow challenged what may or may not have been sent during past measurements by ZGrab. Also, can we easily identify what parts of a configuration would be overwritten by the addition of a `ClientHelloConfiguration`? Providing the raw ClientHello with the scans is an almost unequivocal way to show what was sent to the server (apart from the SNI), and I believe it would be way more sound.

zakird:
I'm inclined to say that this feature can coexist with #224 and given this explanation, probably makes sense to. I'm not sure if this is the right approach, and I'll defer to David on that for the time being, but I'm not opposed to this being integrated into ZGrab.

dadrian:
@mtury I had forgotten about the `unknownExtensions` field. I'm now inclined to agree with @zakird that we should simply merge both.

main.go
```diff
@@ -100,6 +101,8 @@ func init() {
 	flag.BoolVar(&config.S7, "s7", false, "Send some Siemens S7 data")
 	flag.BoolVar(&config.NoSNI, "no-sni", false, "Do not send domain name in TLS handshake regardless of whether known")
 
+	flag.StringVar(&clientHelloFileName, "raw-client-hello", "", "Provide a raw ClientHello to be sent; only the SNI will be rewritten (implies --tls)")
```
dadrian: I don't think this actually implies `--tls` as written, which is fine. I'd just remove the text here.

ztools/ztls/common.go
```diff
@@ -352,6 +352,10 @@ type Config struct {
 
 	// Explicitly set Client random
 	ClientRandom []byte
+
+	// Explicitly set ClientHello
+	ForceClientHello bool
```
dadrian: I think we can get rid of this bool, and just use`ExternalClientHello != nil` as an indicator that we have an external client hello?

ztools/ztls/handshake_messages.go
```diff
@@ -545,7 +570,8 @@ func (m *serverHelloMsg) equal(i interface{}) bool {
 		m.ocspStapling == m1.ocspStapling &&
 		m.ticketSupported == m1.ticketSupported &&
 		m.secureRenegotiation == m1.secureRenegotiation &&
-		m.extendedMasterSecret == m1.extendedMasterSecret
+		m.extendedMasterSecret == m1.extendedMasterSecret &&
+		reflect.DeepEqual(m.unknownExtensions, m1.unknownExtensions)
```
dadrian: This can probably just be bytes.Equal and not the full-blown reflect.DeepEqual

ztools/ztls/common.go
mtury: You're right. I removed the bool in the latest commit!

ztools/ztls/handshake_messages.go
mtury: `unknownExtensions` is a 2D slice (every extension has its own slice of bytes), so unfortunately bytes.Equal cannot be used here.

dadrian:
@mtury I think this PR is good, but it apparently conflicts with master. Can you update it so it merges clean?

mtury:
That was due to the recent PR for ALPN extensions. It should be fine now!

dadrian:
Thanks for your contribution!

PR #224, opened 2017-02-03
Add config field to enable custom client fingerprints

bvandersloot:
These changes enable a zTLS consumer to specify exactly the format of the ClientHello message. 
This is useful for censorship-resistance, so zTLS clients can camouflage themselves as other traffic.

The mechanism this uses is an additional field in the Config struct, ClientFingerprint, that when non-null specifies the contents of the ClientHello message. When this field is nil, everything proceeds as normal. 

I repeat: **WHEN THIS CONFIG FIELD IS nil, THESE CODE PATHS ARE NOT HIT**

The extensions supported out-of-the-box are:

- SNI
- ExtendedMasterSecret
- Renegotiation Info
- Elliptic Curves
- EC point formats
- SessionTickets
- NPN
- StatusRequests
- Signature and Hash Algorithms
- ALPN (pending #220 and some minor changes)

But a consumer can also specify their own extensions and roll them in as well.
Custom Session Caches are also allowed.


zakird:
This seems very similar to #223. Do we need both approaches?

bvandersloot:
We don't need both. Mine is only a modification to the zTLS library, and will not be accessible from the zGrab command line without further changes. My version also does not support sending arbitrary (potentially unimplemented, or even Google black-box) extensions over the wire out-of-the-box.

dadrian:
I'm currently more inclined to this approach than #223. That being said, let's move any broad discussion of goals, etc to #223, rather than having to split threads. If we end up with your approach, I'll come back and review the code here.

dadrian:
@bvandersloot Can you resolve merge conflicts with #223? Then I'll review.

bvandersloot:
Sad!

ztools/ztls/common.go
```diff
@@ -226,6 +226,10 @@ type ClientSessionCache interface {
 	Put(sessionKey string, cs *ClientSessionState)
 }
 
+type ClientExtension interface {
+	Marshal() (data []byte)
```
dadrian: You don't need a named result here.

ztools/ztls/common.go
```diff
@@ -356,6 +360,10 @@ type Config struct {
 
 	// Explicitly set ClientHello with raw data
 	ExternalClientHello []byte
+
+	// If non-null specifies the contents of the client-hello
+	// WARNING: Setting this may invalidate other fields in the Config object
+	ClientFingerprint *ClientHelloConfiguration
```
dadrian: Why are these named differently?

ztools/ztls/handshake_client.go
```diff
@@ -32,19 +32,246 @@ type clientHandshakeState struct {
 	session         *ClientSessionState
 }
 
+type CacheKeyGenerator interface {
+	Key(net.Addr) string
+}
+
+type ClientHelloConfiguration struct {
+	//Version in the handshake header
```
dadrian: There should be a space between `//` and the rest of the comment.

ztools/ztls/handshake_client.go
```diff
@@ -32,19 +32,246 @@ type clientHandshakeState struct {
...
+	//Optional, both must be non-nil, or neither.
+	//Custom Session cache implementations allowed
+	SessionCache ClientSessionCache
+	CacheKey     CacheKeyGenerator
+}
+
+func (c *ClientHelloConfiguration) ValidateExtensions() error {
+	for _, ext := range c.Extensions {
+		switch ext.(type) {
+		case PointFormatExtension:
```
dadrian: Why are there only two extensions here? What happens in the default case?

bvandersloot:
@dadrian: This should be ready for eyes.

ztools/ztls/handshake_client.go
```diff
@@ -32,19 +32,246 @@ type clientHandshakeState struct {
...
+				if !found {
+					return errors.New(fmt.Sprintf("Unsupported Hash and Signature Algorithm (%d, %d)", algs.hash, algs.signature))
+				}
+			}
+		}
+	}
+	return nil
+}
+
+func (c *ClientHelloConfiguration) ModifyConfig(config *Config) *Config {
```
dadrian: I'm not sure I like this name. Maybe `WriteToConfig`?

ztools/ztls/handshake_client.go
```diff
@@ -32,19 +32,246 @@ type clientHandshakeState struct {
...
+	config.NextProtos = []string{}
+	config.CipherSuites = c.CipherSuites
+	config.MaxVersion = c.HandshakeVersion
+	config.ClientRandom = c.ClientRandom
+	config.CurvePreferences = []CurveID{}
+	config.HeartbeatEnabled = false
+	config.ExtendedRandom = false
+	config.ForceSessionTicketExt = false
+	config.ExtendedMasterSecret = false
+	config.SignedCertificateTimestampExt = false
```
dadrian: Safer to do `config = &Config{}`.

ztools/ztls/handshake_client.go
```diff
@@ -32,19 +32,246 @@ type clientHandshakeState struct {
...
+	for i, ext := range c.Extensions {
+		switch ext.(type) {
```
dadrian: This does not have a default case

ztools/ztls/handshake_client.go
```diff
@@ -32,19 +32,246 @@ type clientHandshakeState struct {
...
+		case SniExtension:
```
dadrian: You should be able to do assignment in the switch statement, rather than having to recast inside of each case body.

ztools/ztls/handshake_client.go
```diff
@@ -32,19 +32,246 @@ type clientHandshakeState struct {
...
+			if ext.(SniExtension).Autopopulate {
```
dadrian: What does autopopulate mean?

ztools/ztls/handshake_client.go
```diff
@@ -32,19 +32,246 @@ type clientHandshakeState struct {
...
+			}
+			if config.ServerName == "" && len(ext.(SniExtension).Domains) > 0 {
+				config.ServerName = ext.(SniExtension).Domains[0]
+			}
+		case ALPNExtension:
+			config.NextProtos = ext.(ALPNExtension).Protocols
+		case ExtendedMasterSecretExtension:
+			config.ExtendedMasterSecret = true
+		case SignatureAlgorithmExtension:
+			supportedSKXSignatureAlgorithms = ext.(SignatureAlgorithmExtension).getStruct()
```
dadrian: getStruct is an incredibly vague function name

ztools/ztls/handshake_client.go
```diff
@@ -32,19 +32,246 @@ type clientHandshakeState struct {
...
+		case SupportedCurvesExtension:
+			config.CurvePreferences = ext.(SupportedCurvesExtension).Curves
+		case SessionTicketExtension:
+			config.ForceSessionTicketExt = true
+		}
+	}
+	return config
+}
+
+func (c *ClientHelloConfiguration) marshal(config *Config) ([]byte, error) {
```
dadrian: Why is this method on `ClientHelloConfiguration`? Why does it need a `Config`?

ztools/ztls/handshake_client.go
bvandersloot: Fixing throughout.

ztools/ztls/handshake_client.go
bvandersloot: Most extensions don't need to make sure their details are implemented, so the default is no action. Validate may not be the best wording. Also, maybe this is best rolled into another function in the Extension interface, and calling it on all extensions here?

ztools/ztls/handshake_client.go
bvandersloot: This only updates fields that will be respecified by the ClientFingerprint, so that would not have the desired behavior.

ztools/ztls/handshake_client.go
bvandersloot: adding explicit default: continue

ztools/ztls/handshake_client.go
bvandersloot: I forgot you can do that!

ztools/ztls/handshake_client.go
bvandersloot: Fill in the SNI extension per-request. That way a new Config is not needed per-domain.

ztools/ztls/handshake_client.go
bvandersloot: :%s/getStruct/getStructuredAlgorithms/g

ztools/ztls/handshake_client.go
bvandersloot: This is making the on-the-wire version of the `ClientHelloConfiguration`. I have this distinct from the `clientHelloMsg` serializer to make clearer that they are very different cases. It needs the `Config` to specify the random number generator, and to see if the config is specifying `ForceSuites`.

ztools/ztls/common.go
bvandersloot: Sure. Changing.

ztools/ztls/handshake_client.go
dadrian: I agree, should probably be part of the extension interface.

ztools/ztls/common.go
bvandersloot: Refactoring both to ClientFingerprintConfiguration

ztools/ztls/handshake_client.go
dadrian: Blegh. This is the downside of having two separate codepaths. I'm somewhat inclined to say screw it, and remove the original codepath so that we only build ClientHello's using the ClientFingerprint approach, but that also makes me nervous for obvious reasons.

My original worry here was that we would add a field to `Config`, and then forget to update this function. Thoughts?

ztools/ztls/common.go
```diff
@@ -226,6 +226,10 @@ type ClientSessionCache interface {
 	Put(sessionKey string, cs *ClientSessionState)
 }
 
+type ClientExtension interface {
```
bvandersloot: This also should not be in this file- moving it to be next to the definition of `ClientFingerprintConfiguration`

ztools/ztls/handshake_client.go
dadrian: So a ClientHelloConfiguration still marshals into a TLS ClientHello?

bvandersloot: Yep.

dadrian:
Sorry about the somewhat vague comments, I'm still trying to work out a mental model of this in my head.

ztools/ztls/handshake_client.go
bvandersloot: This is a reasonable original worry. I'm a bigger fan of the dual codepaths. The failure case here is that the config thinks something is supported, but doesn't tell the server. That likely won't cause issues because a server just won't ack the extension in the ServerHello. And if it does, it is a config-change away from being fixed. This code is just trying to handle dumb config inconsistencies gracefully.

Alternatively, shifting the config format significantly sounds like a major pain to every consumer, and rife with potential major errors. 

2 > 1 IMO

bvandersloot:
No problem, there is some complexity here.

bvandersloot:
This is ready for another pass @dadrian

ztools/ztls/handshake_client.go
```diff
@@ -32,19 +32,235 @@ type clientHandshakeState struct {
...
+	CompressionMethods []uint8
+	Extensions         []ClientExtension
+
+	// Optional, both must be non-nil, or neither.
+	// Custom Session cache implementations allowed
+	SessionCache ClientSessionCache
+	CacheKey     CacheKeyGenerator
+}
+
+type ClientExtension interface {
```
dadrian: Can you document how this interface is supposed to behave?

ztools/ztls/handshake_client.go
```diff
@@ -32,19 +32,235 @@ type clientHandshakeState struct {
...
+	config.HeartbeatEnabled = false
+	config.ExtendedRandom = false
+	config.ForceSessionTicketExt = false
+	config.ExtendedMasterSecret = false
+	config.SignedCertificateTimestampExt = false
+	for i, ext := range c.Extensions {
+		switch casted := ext.(type) {
+		case SniExtension:
+			if casted.Autopopulate {
+				c.Extensions[i] = SniExtension{[]string{config.ServerName}, true}
```
dadrian: You should use named fields here, instead of relying on ordering. This also prevents compilation from breaking if you add a field.

```
c.Extensions[i] = SniExtension{
   Domains: []string{config.ServerName},
   Autopopulate: true,
}

File: ztools/ztls/handshake_client.go [hunk 15]
```diff
@@ -32,19 +32,235 @@ type clientHandshakeState struct {
...
+			}
+			if config.ServerName == "" && len(casted.Domains) > 0 {
```
dadrian: I'm a little confused by this logic. We only use `casted.Domains` if config.ServerName is empty? Can you add some comments to explain what's going on?

ztools/ztls/handshake_client.go
```diff
@@ -32,19 +32,235 @@ type clientHandshakeState struct {
...
+	config.MaxVersion = c.HandshakeVersion
+	config.ClientRandom = c.ClientRandom
+	config.CurvePreferences = []CurveID{}
+	config.HeartbeatEnabled = false
+	config.ExtendedRandom = false
+	config.ForceSessionTicketExt = false
+	config.ExtendedMasterSecret = false
+	config.SignedCertificateTimestampExt = false
+	for i, ext := range c.Extensions {
+		switch casted := ext.(type) {
```
dadrian: This entire switch should be made a method in the ClientExtension interface.
```
WriteToConfig(*Config) error
```

ztools/ztls/handshake_client.go
```diff
@@ -32,19 +32,235 @@ type clientHandshakeState struct {
...
+	head := make([]byte, 38)
+	head[0] = 1
+	head[4] = uint8(c.HandshakeVersion >> 8)
+	head[5] = uint8(c.HandshakeVersion)
+	if len(c.ClientRandom) == 32 {
+		copy(head[6:38], c.ClientRandom[0:32])
+	} else {
+		_, err := io.ReadFull(config.rand(), head[6:38])
+		if err != nil {
+			return nil, errors.New("tls: short read from Rand: " + err.Error())
```
dadrian: I'd probably use `fmt.Errorf` for this, but /shrug

ztools/ztls/handshake_client.go
```diff
@@ -32,19 +32,235 @@ type clientHandshakeState struct {
...
+		}
+	}
+
+	sessionID := make([]byte, len(c.SessionID)+1)
```
dadrian: Has `len(c.SessionID)` been confirmed to be less than 256 yet?

ztools/ztls/handshake_client.go
```diff
@@ -32,19 +32,235 @@ type clientHandshakeState struct {
...
+	sessionID[0] = uint8(len(c.SessionID))
+	if len(c.SessionID) > 0 {
+		copy(sessionID[1:], c.SessionID)
+	}
+
+	ciphers := make([]byte, 2+2*len(c.CipherSuites))
+	ciphers[0] = uint8(len(c.CipherSuites) >> 7)
+	ciphers[1] = uint8(len(c.CipherSuites) << 1)
```
dadrian: Wat. Really?

ztools/ztls/handshake_client.go
```diff
@@ -32,19 +32,235 @@ type clientHandshakeState struct {
...
+	for i, suite := range c.CipherSuites {
+		if !config.ForceSuites {
```
dadrian: What does ForceSuites do?

dadrian: I'm pretty sure that I added this pre-FREAK attack, and it's now safe to remove.

ztools/ztls/handshake_client.go
```diff
@@ -32,19 +32,235 @@ type clientHandshakeState struct {
...
+			if !found {
+				return nil, errors.New(fmt.Sprintf("tls: unimplemented cipher suite %d", suite))
+			}
+		}
+
+		ciphers[2+i*2] = uint8(suite >> 8)
+		ciphers[3+i*2] = uint8(suite)
+	}
+
+	compressions := make([]byte, len(c.CompressionMethods)+1)
```
dadrian: Same comment about length verification

ztools/ztls/handshake_client.go
```diff
@@ -32,19 +32,235 @@ type clientHandshakeState struct {
...
+	compressions[0] = uint8(len(c.CompressionMethods))
+	if len(c.CompressionMethods) > 0 {
+		copy(compressions[1:], c.CompressionMethods)
+		if c.CompressionMethods[0] != 0 {
+			return nil, errors.New(fmt.Sprintf("tls: unimplemented compression method %d", c.CompressionMethods[0]))
+		} else if len(c.CompressionMethods) > 1 {
```
dadrian: This shouldn't be an `else`. It should be a separate `if`

ztools/ztls/handshake_client.go
```diff
@@ -32,19 +32,235 @@ type clientHandshakeState struct {
...
+			return nil, errors.New(fmt.Sprintf("tls: unimplemented compression method %d", c.CompressionMethods[1]))
+		}
+	} else {
+		return nil, errors.New("tls: no compression method")
+	}
+
+	var extensions []byte
+	for _, ext := range c.Extensions {
+		extensions = append(extensions, ext.Marshal()...)
```
dadrian: Can `ext.Marshal()` ever fail?

ztools/ztls/handshake_client.go
```diff
@@ -32,19 +32,235 @@ type clientHandshakeState struct {
...
+	}
+	if len(extensions) > 0 {
+		length := make([]byte, 2)
+		length[0] = uint8(len(extensions) >> 8)
+		length[1] = uint8(len(extensions))
+		extensions = append(length, extensions...)
+	}
+	hello := append(head, append(sessionID, append(ciphers, append(compressions, extensions...)...)...)...)
```
dadrian: This is kind of comical.

ztools/ztls/handshake_client.go
```diff
@@ -32,19 +32,235 @@ type clientHandshakeState struct {
...
+	hello[1] = uint8((len(hello) - 4) >> 16)
```
dadrian: Save `len(hello) - 4` to a variable first

ztools/ztls/handshake_client.go
bvandersloot: Adding some comments for each member function

ztools/ztls/handshake_client.go
bvandersloot: Adding. Calling for each `ext`.

ztools/ztls/handshake_client.go
bvandersloot: Good idea.

ztools/ztls/handshake_client.go
bvandersloot: Adding a comment to clarify. This is to allow certificate verification to proceed using a SNI name.

ztools/ztls/handshake_client.go
bvandersloot: Adding check.

ztools/ztls/handshake_client.go
bvandersloot: Borrowed from handshake_messages.go:157 

This is storing the lower 16 bits of 2x the length of the list of cipher suites.

ztools/ztls/handshake_client.go
bvandersloot: ForceSuites tells the client it is okay to send unsupported cipher suites.

ztools/ztls/handshake_client.go
bvandersloot: Shouldn't be a problem with the checks below, but adding because safety is best.

ztools/ztls/handshake_client.go
bvandersloot: Does the RFC specify to not send a compression method more than once?

ztools/ztls/handshake_client.go
bvandersloot: Nope!

ztools/ztls/handshake_client.go
bvandersloot: Agreed. Do you know a better way to append a bunch of slices together in golang?

ztools/ztls/handshake_client.go
bvandersloot: Adding. Also adding a check that `len(hello) - 4 <= 1 << 24`

bvandersloot:
Comments by @dadrian  up to this point should be addressed by `37bbabb`

ztools/ztls/handshake_client.go
dadrian: Not really. Although, preallocating a slice and then appending in a loop of a [][]byte might read a little easier and be faster? Alternatively, you could try to preallocate everything before you serialize, and then just slice off the relevant chunks as you go.

ztools/ztls/handshake_client.go
dadrian: bold-strategy-cotton

ztools/ztls/handshake_client.go
dadrian: So? The network is evil and wants to kill you.

ztools/ztls/handshake_client.go
dadrian: 👍

ztools/ztls/handshake_client.go
dadrian: Ah, you don't have to bother to support that. It's an outdated option that was used before we actually spoke export ciphers, etc.

dadrian: Well, I guess leave it for now.

dadrian: I filed #236

ztools/ztls/handshake_client.go
dadrian: Can you add a comment? That's clever, but non-obvious. agl++

ztools/ztls/handshake_client.go
dadrian: :+1:

ztools/ztls/handshake_client.go
bvandersloot: Fixed in `a28b378`

ztools/ztls/handshake_client.go
bvandersloot: Adding in `37a4e7f`

ztools/ztls/handshake_client.go
```diff
@@ -40,9 +42,11 @@ type ClientFingerprintConfiguration struct {
 	// Version in the handshake header
 	HandshakeVersion uint16
 
-	// if len == 32, it will specify the client random
+	// if len == 32, it will specify the client random.
 	// Otherwise, the field will be random
-	ClientRandom []byte
+	// except the top 4 bytes if InsertTimestamp is true
+	ClientRandom    []byte
+	InsertTimestamp bool
```
bvandersloot: This seems like a good thing to do, as far as camouflaging clients go. Android does this, apparently.

bvandersloot:
@dadrian Seems like we are converging. I fixed a few things and added one small feature.

ztools/ztls/handshake_client.go
```diff
@@ -32,19 +34,255 @@ type clientHandshakeState struct {
...
+					if id == candidateSession.cipherSuite {
+						cipherSuiteOk = true
+						break
+					}
+				}
+				versOk := candidateSession.vers >= c.config.minVersion() &&
+					candidateSession.vers <= c.config.ClientFingerprintConfiguration.HandshakeVersion
+				if versOk && cipherSuiteOk {
+					session = candidateSession
+				}
```
dadrian: Is this not an error case otherwise? Do you have to match version to the previous session if resuming?

ztools/ztls/handshake_client.go
```diff
@@ -32,19 +34,255 @@ type clientHandshakeState struct {
...
+			}
+		}
+		for i, ext := range c.config.ClientFingerprintConfiguration.Extensions {
+			switch ext.(type) {
```
dadrian: This should probably also be part of the extension interface

ztools/ztls/handshake_client.go
```diff
@@ -32,19 +34,255 @@ type clientHandshakeState struct {
...
+			}
+		}
+		var err error
+		helloBytes, err = c.config.ClientFingerprintConfiguration.marshal(c.config)
+		if err != nil {
+			return err
+		}
+		hello = &clientHelloMsg{}
+		if ok := hello.unmarshal(helloBytes); !ok {
+			return errors.New("Incompatible Custom Client Fingerprint")
```
dadrian: Lowercase `incompatible`

ztools/ztls/handshake_client.go
```diff
@@ -32,19 +32,233 @@ type clientHandshakeState struct {
...
+	SessionCache ClientSessionCache
+	CacheKey     CacheKeyGenerator
+}
+
+type ClientExtension interface {
+	// Produce the bytes on the wire for this extension, type and length included
+	Marshal() []byte
+
+	// Function will return an error if zTLS does not implement the necessary features for this extension
+	CheckImplemented() error
```
bvandersloot: In the Hash and Signature Algorithm Extension, if a algorithm combination listed to be offered by the client is not actually in zTLS. Or even a non-defined value.

ztools/ztls/handshake_client.go
```diff
@@ -32,19 +34,255 @@ type clientHandshakeState struct {
...
-	// first, let's check if a ClientHello template was provided by the user
-	if c.config.ExternalClientHello != nil {
+	if c.config.ClientFingerprintConfiguration != nil {
+		if err := c.config.ClientFingerprintConfiguration.WriteToConfig(c.config); err != nil {
+			return err
+		}
+	}
+
+	// first, let's check if a ClientFingerprintConfiguration template was provided by the config
+	if c.config.ClientFingerprintConfiguration != nil {
```
bvandersloot: Merging the above if into this.

ztools/ztls/handshake_client.go
bvandersloot: This is a special case only for session management.

bvandersloot:
No worries, I've had about the same latency between interactions.

ztools/ztls/handshake_client.go
dadrian: Oh, I understand now.

ztools/ztls/handshake_extensions.go
```diff
@@ -0,0 +1,378 @@
...
+
+func (e NullExtension) CheckImplemented() error {
+	return nil
+}
+
+func (e NullExtension) Marshal() []byte {
+	return []byte{}
+}
+
+type SniExtension struct {
```
dadrian: Should be `SNIExtension`

ztools/ztls/handshake_extensions.go
```diff
@@ -0,0 +1,378 @@
...
+	Curves []CurveID
+}
+
+func (e SupportedCurvesExtension) WriteToConfig(c *Config) error {
+	c.CurvePreferences = e.Curves
+	return nil
+}
+
+func (e SupportedCurvesExtension) CheckImplemented() error {
+	return nil
```
dadrian: We support all possible curves?

ztools/ztls/handshake_extensions.go
```diff
@@ -0,0 +1,378 @@
...
+}
+
+func (e PointFormatExtension) WriteToConfig(c *Config) error {
+	return nil
+}
+
+func (e PointFormatExtension) CheckImplemented() error {
+	for _, format := range e.Formats {
+		if format != pointFormatUncompressed {
+			return errors.New(fmt.Sprintf("Unsupported EC Point Format %d", format))
```
dadrian: replace with `fmt.Errorf`

ztools/ztls/handshake_extensions.go
```diff
@@ -0,0 +1,378 @@
...
+
+type HeartbeatExtension struct {
+	Mode byte
+}
+
+func (e HeartbeatExtension) WriteToConfig(c *Config) error {
+	return nil
+}
+
+func (e HeartbeatExtension) CheckImplemented() error {
```
dadrian: File an issue that we don't actually support Heartbeat, only kind of.

dadrian: We can't receive arbitrary Heartbeat messages at arbitrary times.

ztools/ztls/handshake_extensions.go
```diff
@@ -0,0 +1,378 @@
+package ztls
+
+import (
+	"errors"
+	"fmt"
+)
+
+type NullExtension struct {
```
dadrian: This file should all be pointer receivers, unless you have a good reason not to.

bvandersloot: Adding.

ztools/ztls/handshake_extensions.go
bvandersloot: Changing.

ztools/ztls/handshake_extensions.go
bvandersloot: Adding

ztools/ztls/handshake_extensions.go
bvandersloot: Changing.

ztools/ztls/handshake_extensions.go
bvandersloot: Filed.

bvandersloot: #237

dadrian:
@zakird Any further comments?

zakird:
This mostly looks fine to me. However, this is a lot of code that touches a lot of things in a lot of places, and I have little confidence from reading it over that it works. 

I'd like to see at least a few unit tests that confirm that what goes into a `ClientFingerprintConfiguration` leads to the expected Client Hello. I suspect you must have done this already somehow to see if any of this worked. Let's just get them into tests.

bvandersloot:
@zakird, I went with one large functional test that imitates a Firefox hello, rather than trying to unit test. This seems to match the granularity of existing tests. It passes and I manually inspected the hexdump test file and all of the configuration changes were reflected on the wire.

PR #229, opened 2017-02-04
Changed default EHLO to be client IP instead of blank string

dadrian:


dadrian:
@ajholland Can you update this so we don't have to check the IP every connection, and instead grab it at launch and have it saved somewhere?

main.go
```diff
@@ -223,6 +223,15 @@ func init() {
 		config.SMTP = true
 	}
 
+	if config.SMTP && !config.EHLO {
+		name, err := os.Hostname()
+		if err != nil {
+			zlog.Errorf("Unable to set hostname")
```
dadrian: nit: This should be `Fatalf`, also  write error messages with leading lowercase. Also, this is unable to _get_ hostname, not set.
```
zlog.Fatalf("unable to get hostname for EHLO: %s", err.Error())
```

ajholland: Makes sense, pushed new commit

PR #234, opened 2017-02-08
Compile zgrab_schema.py in TravisCI

dadrian:


zakird:
I'm fine with this. It might make sense to actually just run a zschema command to make sure all the types are valid, etc.

dadrian:
@zakird I wasn't sure which zschema command to run. There's several schemas defined in the file, and the `validate` subcommand requires a JSON file to validate.

PR #239, opened 2017-02-22
Made SMTP, IMAP, and POP3 RFC Compliant

### LINKED ISSUE #238: Send QUIT during SMTP scans.
Also, determine if this is needed for IMAP and POP scans.

ajholland:
Resolves #238 

SMTP server requires Quit command as stated in RFC 5321 Section 3.1

IMAP and POP3 are both ambiguous on whether a server requires a close/quit command because a timeout could be implemented by the server. Either way, sending a quit command probably doesn't hurt.

### DIFF
--- a/zlib/grabber.go
+++ b/zlib/grabber.go
@@ -514,21 +514,20 @@ func makeGrabber(config *Config) func(*Conn) error {
 			}
 		}
 
-		if config.Banners {
-			if config.SMTP {
-				if err := c.SMTPQuit(); err != nil {
-					return err
-				}
-			} else if config.POP3 {
-				if err := c.POP3Quit(); err != nil {
-					return err
-				}
-			} else if config.IMAP {
-				if err := c.IMAPQuit(); err != nil {
-					return err
-				}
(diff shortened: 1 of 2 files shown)


zlib/grabber.go
```diff
@@ -514,6 +514,21 @@ func makeGrabber(config *Config) func(*Conn) error {
 			}
 		}
 
+		if config.Banners {
```
zakird: Why is this dependent on banners?

ajholland: Fixed in latest commit, I didn't fully understand banners.

zlib/grabber.go
```diff
@@ -514,6 +514,20 @@ func makeGrabber(config *Config) func(*Conn) error {
 			}
 		}
 
+		if config.SMTP {
+			if err := c.SMTPQuit(); err != nil {
```
dadrian: Can we set `erroredComponent` here in all of the cases, probably to `"quit"`?

ajholland: Yup, pushed

PR #241, opened 2017-02-25
Add logging to ztls when acting as the server

aaspring:
Branch name is inaccurate b/c scope changed. ClientHello is now logged, but everything else is too.


ztools/ztls/common.go
```diff
@@ -102,12 +104,81 @@ const (
...
+	}
+	if expectedName := nameForCurve(aux.Value); expectedName != aux.Name {
+		return fmt.Errorf("mismatched curve and name, curve: %d, name: %s, expected name: %s", aux.Value, aux.Name, expectedName)
+	}
+	*curveID = CurveID(aux.Value)
+	return nil
+}
+
+type PointFormat uint8
+
```
dadrian: Did you make this change? Didn't Zakir implement this a while ago? Is something up with this merge? Are you rebased onto the most recent master?

aaspring: I added this. I did not see the logging anywhere. @zakird??

dadrian: Ok, I guess I'm just crazy.

zakird:
This looks fine to me.

ztools/ztls/common.go
zakird: I don't remember ever adding something like this.

PR #246, opened 2017-02-28
Add JSON marshaling to ClientHelloInfo.

aaspring:


ztools/ztls/common.go
```diff
@@ -949,3 +982,157 @@ type ClientHelloInfo struct {
 	// connection to fail.
 	Conn net.Conn
 }
+
+func (info *ClientHelloInfo) MarshalJSON() ([]byte, error) {
+	aux := struct {
```
zakird: so it might make sense to split out this JSON to match the shape of an actual client hello. I realize that Golang is dumb and makes everything flat, but it'd be nice to at least see extensions in something called `extensions`.

zakird: Also, how are you handling unknown client hello extensions?

ztools/ztls/common.go
```diff
@@ -949,3 +982,157 @@ type ClientHelloInfo struct {
...
+	for i, format := range info.SupportedPoints {
+		aux.SupportedPoints[i] = PointFormat(format)
+	}
+
+	aux.SupportedVersions = make([]TLSVersion, len(info.SupportedVersions))
+	for i, version := range info.SupportedVersions {
+		aux.SupportedVersions[i] = TLSVersion(version)
+	}
+
+	aux.LocalAddr = fmt.Sprintf("%s+%s", info.Conn.LocalAddr().String(), info.Conn.LocalAddr().Network())
```
zakird: This isn't really Client Hello, is it?

ztools/ztls/common.go
```diff
@@ -949,3 +982,157 @@ type ClientHelloInfo struct {
...
+		CipherSuites      []CipherSuite     `json:"cipher_suites"`
+		ServerName        string            `json:"server_name,omitempty"`
+		SupportedCurves   []CurveID         `json:"supported_curves,omitempty"`
+		SupportedPoints   []PointFormat     `json:"supported_point_formats,omitempty"`
+		SignatureSchemes  []SignatureScheme `json:"signature_schemes,omitempty"`
+		SupportedProtos   []string          `json:"supported_protocols,omitempty"`
+		SupportedVersions []TLSVersion      `json:"supported_versions,omitempty"`
+		LocalAddr         string            `json:"local_address,omitempty"`
+		RemoteAddr        string            `json:"remote_address,omitempty"`
+	}{}
```
zakird: This code look _very_ similar to what Ben wrote the other day.

ztools/ztls/common.go
aaspring: No this is a trimmed down version exposed by Golang 1.8 and not the entire Client Hello.

ztools/ztls/common.go
aaspring: @bvandersloot is this a code-clone of somewhere I missed?

zakird:
Given that this is just ported 1.8 stuff, I'm fine with it.

zakird:
@aaspring this looks fine but conflicts with master.

ztools/ztls/ztls_names.go
```diff
@@ -435,6 +436,21 @@ func init() {
 	pointFormatNames[0] = "uncompressed"
 	pointFormatNames[1] = "ansiX962_compressed_prime"
 	pointFormatNames[2] = "ansiX962_compressed_char2"
+
+	// https://tools.ietf.org/html/draft-ietf-tls-tls13-18#section-4.2.3
```
dadrian: Can we make the integers themselves named constants, then define the names map in terms of the constants? In the past, I've just autogenerated this code using awk, then copy/pasted it.

aaspring: Currently there is only a pointFormatUncompressed defined as a uint8 constant in common.go. Do you mean just add constants for the other uint8 values and then change the key to the constant instead of the value?

aaspring: Nevermind, you're talking about the signatureSchemeNames map.

PR #253, opened 2017-03-04
Add ExplicitCurvePreferences to allow no supported ECC curves.

aaspring:
Currently, all servers are required to support ECC curves for all handshakes (even if non-ECDHE). If CurvePreferences is left empty, then it will default to CurveP256, CurveP384, and CurveP521. This is misleading when dealing with configurations where curves were intentionally left unset, but 3 are supported.

### DIFF
--- a/ztools/ztls/common.go
+++ b/ztools/ztls/common.go
@@ -418,6 +418,10 @@ type Config struct {
 	// be used.
 	CurvePreferences []CurveID
 
+	// If enabled, empty CurvePreferences indicates that there are no curves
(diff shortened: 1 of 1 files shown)


zakird:
We're working on moving ztls into its own repo. I believe the copy out has already occurred at this point, and we just need to delete the version in ZGrab this afternoon. Would you be willing to move this PR to https://github.com/zmap/ztls?

zakird:
I'm closing this. Please coordinate with @ajholland and submit to https://github.com/zmap/zcrypto.

Issue #243, opened 2017-02-26
ZTLS should probably be it's own project


dadrian:
ZTLS is big and used outside of scanning for various research-y applications. It doesn't entirely make sense to be part of ZGrab.

ajholland:
Resolved in #258

PR #254, opened 2017-03-04
Remove ztls, update references

ajholland:
Following from #243 I moved ztls to zmap/ztls, removed it from zgrab, and updated the references in existing files


zakird:
This looks fine offhand, but I'd like to get tests running in the ztls repo running before we make the switch. Can you add a .travis.yml there? If possible let's try to get this merged today to reduce confusion of having code in two places. @dadrian can you also take a look?

Zakir

ajholland:
Added .travis.yml but I don't have sufficient permissions to add the repo to travis

I was going to put the tests/code in folders but it turned out to be a little more complicated with the way go does tests and the way we had paths specified. If you want it done I can do it. 

Otherwise this should be good.

PR #255, opened 2017-03-05
Convert Extended Key Usage JSON to a struct

dadrian:


ztools/x509/extensions.go
```diff
@@ -67,19 +71,19 @@ type BasicConstraints struct {
 }
 
 type NoticeReference struct {
```
dadrian: ಠ_ಠ @ whoever made this originally

zakird:
lgtm

PR #259, opened 2017-03-07
Fix nil pointer dereference in big.Int.Cmp().

aaspring:


dadrian:
@aaspring Can you do this against https://github.com/zmap/zcrypto instead?

aaspring:
PR already exists for zmap/zcrypto repo (PR 6). @zakird wanted patches to both until zgrab is completely switched to zcrypto.

Issue #187, opened 2016-09-29
HTTP Memory Leak


zakird:
It looks like ZGrab leaks memory when performing HTTP scans. This is bad enough that it leaks enough memory that it is OOM killed on a 16GB machine before an IPv4 scan can be completed. 

The last successful scan we had was on September 5th, so it may be that the recent port following changes were what caused this.

micgob:
HI,
Any workarounds?

mawenlong3:
ZGrab has also the problem with  leaks memory when performing HTTP scans. I used the command like this 
`zmap -p 80 -o - -B 30M -l 80-http-ipv4_full-20170426T1630-zmap-log.log -m 80-http-ipv4_full-20170426T1630-zmap-metadata.json| ztee 80-http-ipv4_full-20170426T1630-zmap-results.csv | zgrab --port=80 --http="/"  --timeout=5 --senders=2000  --http-max-redirects=2 --output-file=80-http-ipv4_full-20170426T1630-zgrab-results.json --metadata-file=80-http-ipv4_full-20170426T1630-zgrab-metadata.json --log-file=80-http-ipv4_full-20170426T1630-zgrab-log.log`
some hours later,I get the error ‘Apr 27 09:11:59.386 [FATAL] ztee: Error writing to stdout’

PR #268, opened 2017-05-11
Refork net/http to resolve memory leak

ajholland:
This PR reforks the golang net/http library to resolve the memory leak in zgrab (#187). This PR also reapplies all necessary modifications to the golang http library for it to interface with zgrab. 

Running an alexa scan after all of the modifications confirms that the memory leak is gone after reforking. 

Notes on this PR: 
- Output now includes extra fields of "body" and "body-sha256" for redirect response chain. These extra fields should be the only difference between the output of zgrab previously and the output from this PR.  
- HTTPv2 is supported in this reforked library however there were compatibility issues with zcrypto/tls and crypto/tls that caused tests to fail. Since it's not the goal of this PR I've commented out relevant httpv2 tests so that we can support that protocol at a later date without reforking.


zakird:
Can someone spell out the changes that this causes to data output?

dadrian:
In the case where you are following redirects:
- Previously, we did not output body and body sha256 for requests [1:n-1]. We only output body and bodysha256 for the final request (requests[n]).
- Now we output it for each request.

dadrian:
@ajholland can you confirm?

dadrian:
Actually it looks like you might just be experiencing `omitempty`, it looks like `body` and `body_sha256` were already in the schema for the redirect chain.

https://github.com/zmap/zgrab/blob/master/zgrab_schema.py#L534

ajholland:
Confirmed that it was just `omitempty`, I updated so that referer is added to list of known headers. `user_agent` is now the only difference between output previously and output with this PR. `user_agent` was previously set after we had parsed headers and therefore wasn't caught by the parse function. It's not in the schema or known headers so this should be fine. 

`user_agent` is now listed as an unknown header object as opposed to being labeled.

dadrian:
LGTM, but still should wait for @zakird to approve.

dadrian:
Actually, @ajholland you still need to run `go fmt`. Consider adding `go fmt` or `goimports` to your text editor to run automatically on save.

ztools/http/chunked.go
```diff
@@ -159,12 +206,46 @@ func (cw *chunkedWriter) Write(data []byte) (n int, err error) {
...
+// FlushAfterChunkWriter signals from the caller of NewChunkedWriter
+// that each chunk should be followed by a flush. It is used by the
+// http.Transport code to keep the buffering behavior for headers and
+// trailers, but flush out chunks aggressively in the middle for
+// request bodies which may be generated slowly. See Issue 6574.
+type FlushAfterChunkWriter struct {
+	*bufio.Writer
+}
+
+func parseHexUint(v []byte) (n uint64, err error) {
```
zakird: Wat.

zakird:
With a PR this large, I'm hesitant to believe that looking through code is going actually going to find problems. I think the best way to proceed is probably through some empirical tests.

I don't know how much of this has been done informally, but would like to see some results from:

- breakdown of statuses/errors for a shared 1% sample of IPv4 and Alexa Top Million.
  - look at what sites have errors in the new copy that don't in the old

- look at some pages that have different hashes or redirect chains and make sure they look right

- run output through `zschema validate` and make sure there aren't any surprises

ztools/http/chunked.go
ajholland: I'm confused, what's the issue here? This is just brought in from the updated net/http.

ajholland:
Sure, I can definitely do that, informally I ran two scans on separate scratch machines at nearly the same time over a thousand domain scan and the result difference was only 1 new error instead of a success, I'd attribute that to unpredictability in timing or packets. 

I'll work on these tests and get the results to you and dadrian.

PR #271, opened 2017-05-18
remove log error

ajholland:
This error arises when a header is incorrectly formatted or "bogus stuff". https://github.com/golang/go/issues/19895
Removing it is good because we encounter incorrectly formatted stuff quite often.


zakird:
😛

PR #274, opened 2017-05-23
Different output for empty responses

ajholland:
Many CDN http servers will return an empty response to an HTTP GET
with an IP host header. In the censys inferface (and from the ipv4
table) you can't distingush between these hosts and a host that has
port 80 closed.

EOF is the error returned and the IP is marked fail, this PR marks the IP as returning empty HTTP and successful IP. 

Note: the test that is currently failing is testing for the behavior that we're removing. 

Is this the type of output we would like? i.e. empty http json 
Does some other part of Censys have to be modified to get these IP's to show up?

### DIFF
--- a/ztools/http/client.go
+++ b/ztools/http/client.go
@@ -613,7 +613,12 @@ func (c *Client) Do(req *Request) (resp *Response, err error) {
 					timeout: true,
 				}
 			}
+			if err == io.EOF {
+				return nil, err
+			}
+
 			return nil, uerr(err)
+
 		}
 
 		var shouldRedirect bool
--- a/ztools/http/client.go
+++ b/ztools/http/client.go
@@ -618,7 +618,6 @@ func (c *Client) Do(req *Request) (resp *Response, err error) {
 			}
 
 			return nil, uerr(err)
-
 		}
 
 		var shouldRedirect bool
--- a/ztools/http/client.go
+++ b/ztools/http/client.go
@@ -613,10 +613,6 @@ func (c *Client) Do(req *Request) (resp *Response, err error) {
 					timeout: true,
 				}
 			}
-			if err == io.EOF {
-				return nil, err
-			}
-
 			return nil, uerr(err)
 		}
 
--- a/zlib/grabber.go
+++ b/zlib/grabber.go
@@ -301,7 +301,7 @@ func makeHTTPGrabber(config *Config, grabData *GrabData) func(string, string, st
 		}
 		grabData.HTTP.Response = resp
 
-		if err.Error() == "EOF" {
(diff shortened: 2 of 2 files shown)


zakird:
@dadrian thoughts?

dadrian:
@ajholland and I discussed this briefly. `io.EOF` is only returned on "gracefully" closed connections to indicate, well, EOF. So this seems reasonable to me.

The other way to fix this in Censys would be to update ztag, but I'd like to see a model where there was less of a difference between a zgrab success and a ztag success, so 👍 from me.

zakird:
FWM them

ztools/http/client.go
```diff
@@ -613,6 +613,10 @@ func (c *Client) Do(req *Request) (resp *Response, err error) {
 					timeout: true,
 				}
 			}
+			if err == io.EOF {
```
dadrian: Wait, do we need this change? The HTTP lib should ideally behave the same, we should catch any differences zgrab has between HTTP at the ZGrab logic. Doesn't the change to `grabber.go` fix that?

ajholland: No, the HTTP lib returns a custom error message that includes the get and the ip address followed by io.EOF. I could parse the string in grabber.go to see if it contains io.EOF but this seemed easier.

dadrian: What error type is actually returned? Can you cast it to it's real type and then check if it contains an io.EOF, or is that not possible?

ajholland: Yup, changed and pushed.

dadrian:
So this results in output like this, as does my cleaned up version which I'll push in a moment:
```console
$ ./zgrab --http="/" --http-max-redirects=0 <<< "36.66.90.248" | jq .
{
  "ip": "36.66.90.248",
  "timestamp": "2017-05-29T13:25:55-04:00",
  "data": {
    "http": {}
  }
}
```

We don't want `"http"` to be empty here. For starters, we prefer omitted to empty. Second, for this to be fully fixed, we want headers as well. The problem is, the HTTP lib helpfully nil's out the response in all error cases, so we lose this information. We should investigate changing this behavior inside of the HTTP library, see how much work that is, and confirm that it doesn't introduce a memory leak.

I'm also open to other solutions, but I don't have any off the top of my head. @ajholland, thoughts?

ajholland:
I agree, that seems like the best solution (changing the behavior of the http library). It shouldn't be a large deal since io.EOF only occurs at the "graceful end of input". I'll make sure to not introduce a memory leak.

zakird:
@ajholland what's up with this PR?

ajholland:
It's on my list of things to fix/do but haven't yet, I'll close it until I get a chance.

PR #277, opened 2017-06-14
Add SMB to zgrab_schema

ajholland:


zgrab_schema.py
```diff
@@ -670,6 +670,16 @@
 
 zschema.registry.register_schema("zgrab-s7", zgrab_s7)
 
+zgrab_smb = Record({
+    "data":SubRecord({
+        "smb":SubRecord({
+           "smbv1":Boolean(),
```
zakird: i might label that last field `smb1_support` to make it clear that it's not a subrecord.

zakird:
This looks fine. You probably also want to fix the code to match this ;)

dadrian:
@cdzombak I believe @ajholland is out this week, can you take over this PR?

cdzombak:
> can you take over this PR?

@dadrian yes. I am not familiar with the goal for this work, though, and I see no open issues on this repo or in Trello. What do you need me to do?

ajholland:
Since this is a change to the schema and most? everything else was already merged I think this just needs to be merged

dadrian:
@cdzombak Zakir's comment still needs to be addressed, as far as I can tell. I may be mistaken?

> This looks fine. You probably also want to fix the code to match this ;)

If I"m understanding correctly, @zakird requested a name in the output be changed, which @ajholland updated in the schema, but the necessary change to the actual JSON ouput in the SMB code needs to be made as well (e.g. changing the `json:blah` struct tag).

cdzombak:
@dadrian makes sense. I'll get to that tomorrow.

cdzombak:
1f3067b makes the necessary name change in the grabber itself.

On my home network, a scan results in the following JSON, which seems to match the schema:

```json
{"ip":"192.168.0.165","timestamp":"2017-06-20T09:38:34-04:00","data":{"smb":{"smbv1_support":false}},"error":"Invalid NetBIOS Session message","error_component":"smb"}
```

PR #278, opened 2017-06-19
Avoid using deprecated types in XSSH schema

cdzombak:
ZSchema warns that the Short and Integer types are deprecated. This changes their usage in the XSSH schema to appropriate non-deprecated types.

### DIFF
--- a/zgrab_schema.py
+++ b/zgrab_schema.py
@@ -773,7 +773,7 @@ zgrab_xssh = Record({
                 "client_to_server_languages":ListOf(String()),
                 "server_to_client_languages":ListOf(String()),
                 "first_kex_follows":Boolean(),
-                "reserved":Short(),
+                "reserved":Unsigned32BitInteger(),
             }),
             "userauth":ListOf(String()),
             "algorithm_selection":SubRecord({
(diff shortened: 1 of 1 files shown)


zgrab_schema.py
```diff
@@ -773,7 +773,7 @@
                 "client_to_server_languages":ListOf(String()),
                 "server_to_client_languages":ListOf(String()),
                 "first_kex_follows":Boolean(),
-                "reserved":Short(),
+                "reserved":Unsigned32BitInteger(),
```
zakird: Does this _actually_ need 32 bits? This results logic results in a 64-bit integer.

zgrab_schema.py
```diff
@@ -820,15 +820,15 @@
                         }),
                         "serial":String(),
                         "cert_type":SubRecord({
-                            "id":Integer(),
+                            "id":Unsigned32BitInteger(),
```
zakird: I'm not quite sure what this field is, so can't really provide a review on that it's typed correctly. I'm trusting that you checked whatever relevant documentation.

zgrab_schema.py
cdzombak: [RFC 4253](https://www.ietf.org/rfc/rfc4253.txt) says this is a reserved `uint32` (see page 16 of the RFC)/

zgrab_schema.py
cdzombak: I'm not quite sure what it is, either, but it's populated from [this `uint32` field](https://github.com/zmap/zgrab/blob/42a47296c0227ba36e9a69ff57e6a585a5b81e4b/ztools/xssh/certs.go#L56), so I think this is an appropriate choice in the schema.

PR #279, opened 2017-06-29
Remove hyphens from XSSH grabber output

cdzombak:


ztools/xssh/server.go
```diff
@@ -91,10 +93,11 @@ func (ext *JsonExtensions) MarshalJSON() ([]byte, error) {
 			}
 		}
 
+		jsonSafeKey := strings.Replace(key, "-", "_", -1)
```
cdzombak: This repetition indicates a clear case for future refactoring.

Issue #280, opened 2017-06-30
HTTP Scanner shouldn't follow redirects to localhost


zakird:
We're scraping our own website. :|

ajholland:
What's the intended behavior we'd like here? Print out a message or an error message if we get a redirect to localhost?

zakird:
This was fixed by #281.

PR #281, opened 2017-07-05
Stop redirect before reaching localhost

ajholland:
Addresses #280 

This stops one hop before localhost. Localhost is seen in the Location field of the output request.


zakird:
We should handle IPv6 localhost as well as the aliases that linux would follow.

dadrian:
@ajholland Is this ready to be reviewed?

main.go
```diff
@@ -140,6 +140,7 @@ func init() {
 	// Flags for SMB scanner
 	flag.BoolVar(&config.SMB.SMB, "smb", false, "Scan for SMB")
 	flag.IntVar(&config.SMB.Protocol, "smb-protocol", 1, "Specify which SMB protocol to scan for")
+	flag.BoolVar(&config.NoLocalhost, "no-redirect-localhost", false, "Follow redirect to localhost")
```
zakird: this is named poorly.

cdzombak: it's also under the "Flags for SMB scanner" comment.

ajholland:
@dadrian yes

@zakir and cdzombak, thanks I'll put those changes in!

main.go
```diff
@@ -84,6 +84,7 @@ func init() {
 	flag.StringVar(&config.HTTP.ProxyDomain, "http-proxy-domain", "", "Send a CONNECT <domain> first")
 	flag.IntVar(&config.HTTP.MaxSize, "http-max-size", 256, "Max kilobytes to read in response to an HTTP request")
 	flag.IntVar(&config.HTTP.MaxRedirects, "http-max-redirects", 0, "Max number of redirects to follow")
+	flag.BoolVar(&config.HTTP.FollowLocalhostRedirects, "follow-localhost-redirects", true, "Follow redirects to localhost")
```
cdzombak: In the docs, I would write "Follow redirects to localhost" so that it's clear to readers which scanner this option applies to.

zlib/grabber.go
```diff
@@ -213,6 +215,13 @@ func containsPort(host string) bool {
 	return strings.LastIndex(host, ":") > strings.LastIndex(host, "]")
 }
 
+func redirectsToLocalhost(host string) bool {
+	if host[:3] == "127" || host[:3] == "::1" || host == "localhost" {
```
cdzombak: What happens if the host is <4 characters long, somehow?

main.go
ajholland: docs? 

Do you mean "Follow HTTP redirects to localhost"?

cdzombak: oh, yes, I do.

main.go
```diff
@@ -84,6 +84,7 @@ func init() {
 	flag.StringVar(&config.HTTP.ProxyDomain, "http-proxy-domain", "", "Send a CONNECT <domain> first")
 	flag.IntVar(&config.HTTP.MaxSize, "http-max-size", 256, "Max kilobytes to read in response to an HTTP request")
 	flag.IntVar(&config.HTTP.MaxRedirects, "http-max-redirects", 0, "Max number of redirects to follow")
+	flag.BoolVar(&config.HTTP.FollowLocalhostRedirects, "follow-localhost-redirects", true, "Follow HTTP redirects to localhost")
```
dadrian: This should default to false

zlib/grabber.go
```diff
@@ -213,6 +215,15 @@ func containsPort(host string) bool {
 	return strings.LastIndex(host, ":") > strings.LastIndex(host, "]")
 }
 
+func redirectsToLocalhost(host string) bool {
+	if len(host) >= 3 {
+		if host[:3] == "127" || host[:3] == "::1" || host == "localhost" {
```
dadrian: You should actually convert to an IP and check against subnet masks.

Also, does this work if the host if `example.com` but it resolves to `127.0.0.1`?

main.go
dadrian: I've changed my mind on this.

zlib/grabber.go
zakird: ☝️

main.go
zakird: I disagree with your disagreement with yourself

dadrian: Look, I thought we'd want to default to not doing this as well, but in retrospect realized we should just really pass `--follow-localhost-redirects=false` in the scan runner.

zlib/grabber.go
ajholland: Two seconds, commit incoming.

ajholland:
Most recent version casts IP's and does lookups, tested it locally by downloading all of the IP's that were scraping scans.io, ran a handler on port 80 to report what connects to localhost, then ran those IP's through zgrab, none reached localhost.

zlib/grabber.go
```diff
@@ -213,6 +215,31 @@ func containsPort(host string) bool {
...
+			addrs, err := net.LookupHost(host)
+			if err != nil {
+				zlog.Error(host)
+			}
+			for i := 0; i < len(addrs); i++ {
+				ip := net.ParseIP(addrs[i])
+				if ip == nil {
+					zlog.Error(addrs[i])
+				}
+				if ip.IsLoopback() || ip.Equal(net.IPv4zero) {
```
dadrian: This will panic() if ip is nil

zlib/grabber.go
```diff
@@ -213,6 +215,31 @@ func containsPort(host string) bool {
...
+			return true
+		} else {
+			addrs, err := net.LookupHost(host)
+			if err != nil {
+				zlog.Error(host)
+			}
+			for i := 0; i < len(addrs); i++ {
+				ip := net.ParseIP(addrs[i])
+				if ip == nil {
+					zlog.Error(addrs[i])
```
dadrian: Probably not worth logging

zlib/grabber.go
```diff
@@ -213,6 +215,31 @@ func containsPort(host string) bool {
...
+		return i.IsLoopback() || i.Equal(net.IPv4zero)
+	} else {
+		if host == "localhost" {
+			return true
+		} else {
+			addrs, err := net.LookupHost(host)
+			if err != nil {
+				zlog.Error(host)
+			}
+			for i := 0; i < len(addrs); i++ {
```
dadrian: Use a range-based for. `for i := range addrs`

PR #283, opened 2017-07-12
Remove unused code

ajholland:


ajholland:
@dadrian @zakir this is done/ready
Would it be worth it to move ztools/{util/, processing/, keys/, zct/, zlog/} to zlib/? The motivation would be that all external protocol relevant code would be in ztools/ under the relevant protocol and all of the others/written by us under zlib/. 

Also, do we want to remove zlog/ and switch to logrus? What about migrating zct to it's own repo?

dadrian:
- I wouldn't bother moving anything around in ZGrab unless it helps ZGrab2. Do whatever you want in ZGrab 2. At the same time, I don't really care if you want to move that over or not.
- I don't like zlog, I'm fine removing it. But again, ZGrab 2. We should standardize on a Logrus output format string though
- zct is already in ZCrypto as `zcrypto/ct`, we should just delete it here.

PR #286, opened 2017-07-17
Break out XSSH key exchanges by type in JSON output 

cdzombak:
The principal goal of this PR is to split up the different types of key exchanges for the XSSH grabber into separate names, to allow defining a consistent schema.

There is a little more massaging & rearranging to do to make this data show up nicely in Censys, but I propose doing that in ZTag. (I should have a follow-up PR there this afternoon.)

### DIFF
--- a/zgrab_schema.py
+++ b/zgrab_schema.py
@@ -761,6 +761,11 @@ xssh_signature = SubRecord({
     "raw":Binary(),
 })
 
+golang_crypto_param = SubRecord({
+    "value":Binary(),
+    "length":Integer()
+})
+
 zgrab_xssh = Record({
     "data":SubRecord({
         "xssh":SubRecord({
@@ -800,28 +805,22 @@ zgrab_xssh = Record({
                     "compression":String(),
                 }),
             }),
-            "dh_key_exchange": SubRecord({
-                "params": SubRecord({
-                    "prime": SubRecord({
-                        "value":Binary(),
-                        "length":Integer()
-                    }),
-                    "generator": SubRecord({
-                        "value":Binary(),
-                        "length":Integer()
-                    }),
-                    "client_public": SubRecord({
-                        "value":Binary(),
-                        "length":Integer()
-                    }),
-                    "client_private": SubRecord({
-                        "value":Binary(),
-                        "length":Integer()
-                    }),
+            "key_exchange": SubRecord({
+                "curve25519sha256_params": SubRecord({
+                    "server_public": Binary(),
+                }),
+                "ecdh_params": SubRecord({
                     "server_public": SubRecord({
-                        "value":Binary(),
-                        "length":Integer()
+                        "x": golang_crypto_param,
+                        "y": golang_crypto_param,
                     }),
+                }),
+                "dh_params": SubRecord({
+                    "prime": golang_crypto_param,
+                    "generator": golang_crypto_param,
+                    "client_public": golang_crypto_param,
+                    "client_private": golang_crypto_param,
+                    "server_public": golang_crypto_param,
                 }),
                 "server_signature":xssh_signature,
                 "server_host_key":SubRecord({
--- a/zgrab_schema.py
+++ b/zgrab_schema.py
@@ -818,8 +818,6 @@ zgrab_xssh = Record({
                 "dh_params": SubRecord({
                     "prime": golang_crypto_param,
                     "generator": golang_crypto_param,
-                    "client_public": golang_crypto_param,
-                    "client_private": golang_crypto_param,
                     "server_public": golang_crypto_param,
                 }),
                 "server_signature":xssh_signature,
--- a/zgrab_schema.py
+++ b/zgrab_schema.py
@@ -806,7 +806,7 @@ zgrab_xssh = Record({
                 }),
             }),
             "key_exchange": SubRecord({
-                "ed25519sha256_params": SubRecord({
+                "curve25519_sha256_params": SubRecord({
                     "server_public": Binary(),
                 }),
                 "ecdh_params": SubRecord({
--- a/zgrab_schema.py
+++ b/zgrab_schema.py
@@ -763,7 +763,7 @@ xssh_signature = SubRecord({
 
 golang_crypto_param = SubRecord({
     "value":Binary(),
-    "length":Integer()
+    "length":Unsigned32BitInteger()
 })
 
 zgrab_xssh = Record({
--- a/zgrab_schema.py
+++ b/zgrab_schema.py
@@ -806,7 +806,7 @@ zgrab_xssh = Record({
                 }),
             }),
             "key_exchange": SubRecord({
-                "curve25519sha256_params": SubRecord({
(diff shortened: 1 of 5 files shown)


ztools/xssh/flag.go
```diff
@@ -147,7 +147,7 @@ func init() {
 		strings.Join(defaultCiphers, ","))
 	flag.Var(&pkgConfig.Ciphers, "xssh-ciphers", ciphersUsage)
 
-	flag.BoolVar(&pkgConfig.Verbose, "xssh-verbose", false, "Output additional information.")
+	flag.BoolVar(&pkgConfig.Verbose, "xssh-verbose", false, "Output additional information, including X/SSH client properties from the SSH handshake.")
```
dadrian: OIC

ztools/xssh/kex.go
```diff
@@ -165,7 +165,7 @@ type dhGroup struct {
 	JsonLog       dhGroupJsonLog
 }
 type dhGroupJsonLog struct {
-	Parameters      *ztoolsKeys.DHParams  `json:"params,omitempty"`
+	Parameters      *ztoolsKeys.DHParams  `json:"dh_params,omitempty"`
```
dadrian: I'm a little confused by how this fits together. I was expecting something along the lines of:
```go
type KeyExchange struct {
    DHParams *DHParams `json:"dh_params,omitempty"`
    ECDHParams *ECDHParams `json:"ecdh_params,omitempty"`
    Ed25519Params *Ed25519Params `json:"ed25519_params,omitempty"`
}

type ServerHostKey struct {
    RSAPublic `json:"rsa_public_key,omitempty"`
    ECDHPublic `json:"ecdh_public_key,omitempty"`
    Ed25519Public `json:"ed25519_public_key,omitmepty"`
}

type SSH struct {
    KeyExchange *KeyExchange `json:"key_exchange,omitempty"`
    ServerHostKey *ServerHostKey `json:"server_host_key,omitempty"`
    ServerSignature *ServerSignature `json:"server_signature,omitempty"`
}
```

Why are key exchange types duplicating fields?

zgrab_schema.py
```diff
@@ -761,6 +761,11 @@
     "raw":Binary(),
 })
 
+golang_crypto_param = SubRecord({
+    "value":Binary(),
+    "length":Integer()
```
cdzombak: The ZGrab schema doesn't appear to contain integer sized or make the `Binary`/`IndexedBinary` distinction, while the ZTag schema _does_ contain that info. I continued with that convention here, though I'm happy to adjust as needed.

zgrab_schema.py
```diff
@@ -800,29 +805,23 @@
...
-                    "client_public": SubRecord({
-                        "value":Binary(),
-                        "length":Integer()
-                    }),
-                    "client_private": SubRecord({
-                        "value":Binary(),
-                        "length":Integer()
-                    }),
+            "key_exchange": SubRecord({
+                "ed25519sha256_params": SubRecord({
```
cdzombak: Not AFAIK. But this key exchange is referred to internally (by the x/ssh code) as `curve25519sha256`, and I just carried it forward here.

zgrab_schema.py
```diff
@@ -800,29 +805,23 @@
...
-                        "value":Binary(),
-                        "length":Integer()
+                        "x": golang_crypto_param,
+                        "y": golang_crypto_param,
                     }),
                 }),
+                "dh_params": SubRecord({
+                    "prime": golang_crypto_param,
+                    "generator": golang_crypto_param,
+                    "client_public": golang_crypto_param,
```
cdzombak: [The code that produces](https://github.com/zmap/zgrab/blob/master/ztools/keys/dhe.go#L45) this output is shared between the XSSH and TLS grabbers. Modifying it would affect TLS output, too. I'm not sure whether that's desired.

zgrab_schema.py
```diff
@@ -800,29 +805,23 @@
...
+                    "client_private": golang_crypto_param,
```
cdzombak: Let's continue this discussion in https://github.com/zmap/zgrab/pull/286#discussion_r127828163.

zgrab_schema.py
cdzombak: For this first pass, I opted _not_ to change that shared code and instead to strip this out as part of https://github.com/zmap/ztag/pull/76.

ztools/xssh/kex.go
cdzombak: It was pretty straightforward to make these changes to the existing code and get the output looking like what we want. I had some trouble completely wrapping my head around exactly which log structures are passed to which parts of the handshake, and further restructuring the logging to look like that suggestion would just take more time.

zgrab_schema.py
dadrian: We should still use the new (sized) types for new types. I think that's just an artifact of when we wrote the ZGrab schema, and it's (relative) unuse in any part of Censys now.

zgrab_schema.py
dadrian: Ah, that's because of the "RFC". https://git.libssh.org/projects/libssh.git/tree/doc/curve25519-sha256@libssh.org.txt

We should change the name to `curve25519_sha256_params`, to match the protocol name. And also, because this is key exchange, not signing, so it's `Curve25519` or `X25519` (a Montgomery curve), as opposed to `Ed25519` (an Edwards curve), which is used for signing.

zgrab_schema.py
dadrian: What I do in TLS is just set it to null after kex finishes, unless an option is passed. Or maybe I always set it to null. Regardless, it sounds like that's what `--xssh-verbose` does.

ztools/xssh/kex.go
dadrian: Alright. We should file an issue noting that all of this is confusing, and a little bit about how it works, and note that structure should probably be reconsidered in some way.

zgrab_schema.py
```diff
@@ -761,6 +761,11 @@
     "raw":Binary(),
 })
 
+golang_crypto_param = SubRecord({
+    "value":Binary(),
```
cdzombak: is `IndexedBinary` usable in the ZGrab schema?

zgrab_schema.py
cdzombak: Addressed in cc73e1e34471ab765fdff7905151fa69c5ce8ef7

zgrab_schema.py
cdzombak: Thank you for the clarification. Addressed in 43bcf80bc9473fc1f3ed08bcd2d605f70dc82b1e

zgrab_schema.py
cdzombak: This is fixed in 0de2c77b4f1e84e8fd790b038d8cf3989799917b — I still don't 100% understand how all this logging code works, but I can confirm that commit removed `client_private` from non-verbose XSSH grabs.

zakird:
@dadrian it looks like @cdzombak has updated per your request. Can you take a second look when you have a moment?

PR #287, opened 2017-07-27
Remove ztools/ssh

dadrian:
Note that the old SSH package had some old crypto tricks and client
impersonation behavior that the new one does not. The good thing is the
new one plays nice with Dropbear, etc by default, whereas the old one
you had to play games to get it to connect successfully due to
compatibility issues with Dropbear and OpenSSH implementations.


ajholland:
Should old ssh be removed from schema?

zakird:
@ajholland +1

dadrian:
🐮

PR #290, opened 2017-08-28
Always log basic data for all grabs

cdzombak:
The XSSH grabber isn't outputting anything to the log, which confuses zsearch (it interprets an empty log file as a failure). The easiest way to fix that is for us to always output log info.

This should fix the apparent SSH scan errors we're seeing now.


cdzombak:
This output looks like:

```
Aug 28 13:26:06.103 [INFO] log: Started grab at 2017-08-28T13:26:06-04:00
Aug 28 13:26:06.106 [INFO] log: Finished grab (0 success; 0 failure) at 2017-08-28T13:26:06-04:00
```

main.go
```diff
@@ -359,9 +359,15 @@ func main() {
 	decoder := zlib.NewGrabTargetDecoder(inputFile, config.LookupDomain)
 	marshaler := zlib.NewGrabMarshaler()
 	worker := zlib.NewGrabWorker(&config)
+
 	start := time.Now()
+	zlog.Infof("Started grab at %s", start.Format(time.RFC3339))
```
zakird: lowercase please!

main.go
```diff
@@ -359,9 +359,15 @@ func main() {
 	decoder := zlib.NewGrabTargetDecoder(inputFile, config.LookupDomain)
 	marshaler := zlib.NewGrabMarshaler()
 	worker := zlib.NewGrabWorker(&config)
+
```
ajholland: Can you remove these newlines? Unless you did them purposefully for readability.

main.go
cdzombak: Addressed in f23d962

main.go
cdzombak: I did that purposefully for readability, but if we don't like them I can revert them.

main.go
cdzombak: (and typo fixed in 9e1d4ad)

cdzombak:
Travis is failing with

```
$ go get ./...
package github.com/zmap/zlint/zlint: cannot find package "github.com/zmap/zlint/zlint" in any of:
	/home/travis/.gimme/versions/go1.8.1.linux.amd64/src/github.com/zmap/zlint/zlint (from $GOROOT)
	/home/travis/gopath/src/github.com/zmap/zlint/zlint (from $GOPATH)
The command "go get ./..." failed and exited with 1 during .
```

ajholland:
@cdzombak That's a side effect of zcertificate importing zlint which just had a structure change. I'm pretty sure we don't use zcertificate and could be removed?

main.go
ajholland: I'll defer to zakir, I'm not a large fan though. ¯\_(ツ)_/¯

zakird: This is all going away in like a month. Not a fan, but it just does not matter right now.

ajholland: :+1:

PR #296, opened 2017-11-07
Option to compress output and log with gzip

dfelsch:
Proved to save a lot of storage in large scans

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,6 +37,8 @@ Usage of ./zgrab:
     	Send Chrome Ordered Cipher Suites
   -chrome-no-dhe-ciphers
     	Send chrome ciphers minus DHE suites
+  -compress
+    	Compress output and log using gzip
   -connections-per-host uint
     	Number of times to connect to each host (results in more output) (default 1)
   -data string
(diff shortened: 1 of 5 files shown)


dadrian:
Hi! Thank you for your contribution!

We've also experienced large gains by compressing scan files. However, rather than maintain support for gzip inside of ZGrab, we already support this using unix pipes. For example, `zgrab --port 443 --tls --log-file zgrab-log.txt --metadata-file zgrab-metadata.json | gzip > zgrab-output.gz` compresses ZGrab output and saves it in `zgrab-output.gz`.

Is there a particular reason you need compression in ZGrab itself, vs piping it to gzip, lz4, etc.?

dfelsch:
I had the case that also the log file got very large and I also wanted to compress that.
Other than that, no particular reason. I developed that patch for our project and wanted to share it.

dadrian:
We're actively working on [ZGrab 2](https://github.com/zmap/zgrab2), which isn't going to have nearly as large of log files. To be honest, the log files in ZGrab are way too verbose, since all the errors effectively written in the output file anyway.

Thanks for the patch, but I think I'm going to close this out.

PR #297, opened 2017-11-15
TLSCertsOnly Option

jcrowgey:
This PR is connected to new functionality proposed for zcryto: https://github.com/zmap/zcrypto/pull/100

Scanning goes faster if you don't continue the TLS handshake after getting the certs, and furthermore if you don't parse the TLS Certs on the same host that scans them.  With this option you can scan more hosts faster (by moving on before completing the TLS handshake) and you can collect certificates, diff them against a store of parsed certificates, and only parse the new ones.

### DIFF
--- a/README.md
+++ b/README.md
@@ -173,7 +173,7 @@ $ zmap -p 443 --output-fields=* | ztee results.csv | zgrab --port 443 --tls --ht
 
 ## Requirements
 
-zgrab requires go version of at least 1.6. Please note that this is newer than the version included in Ubuntu 14.04 apt repository. You can install ztee from ZMap Github repository at https://github.com/zmap/zmap.
+zgrab requires go version of at least 1.8.1. Please note that this is newer than the version included in Ubuntu 14.04 apt repository. You can install ztee from ZMap Github repository at https://github.com/zmap/zmap.
 
 
 ## ZGrab as a library / dependency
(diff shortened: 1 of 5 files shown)


jcrowgey:
Note: CI fails because the PR requires the proposed changes to zmap/zcrypto.

jcrowgey:
I've updated this PR to use the listen for the ErrCertsOnly when using the CertsOnly option from the zcrypto/tls package.

dadrian:
@jcrowgey I reran the build now that ZCrypto is updated, but it looks like there's some other incompatibility with `InsecureSkipVerify`. Can you investigate?

jcrowgey:
@dadrian I believe the build is failing because I updated this PR to use the `InsecureSkipValidation` option that I propose in https://github.com/zmap/zcrypto/pull/103.

I see now that it would probably be better to make that a separate PR.

I'll revise this PR to *only* implement `--tls-certs-only` since that functionality is now in ZCrypto.  Then I'll make a separate PR for ZGrab connected to ZCrypto#103 (finish handshake but skip validation).

jcrowgey:
I've rolled back the commits related to zmap/zcrypto#103, I'm going to try closing and reopening this PR to trigger a travis rebuild (https://github.com/travis-ci/travis-ci/issues/576#issuecomment-65317915 ).

PR #299, opened 2017-11-24
Fix domain bug

### LINKED ISSUE #298: --lookup-domain doesn't work
With the command

zgrab --input-file=domains.all --lookup-domain --port 80 --output-file=domains.all.80.callback

when domains.all is a file with one domain per line, doesn't work. The error is produced:

Nov 18 21:59:33.323 [ERROR] banner-grab: Could not connect to <redact> remote host <nil>: dial tcp: lookup <nil>: no such host

Whats the proper usage of --lookup-domain ? --help says input contains domains only.

ajholland:
Resolves #298

Thoughts @dadrian?

### DIFF
--- a/zlib/grabber.go
+++ b/zlib/grabber.go
@@ -644,7 +644,12 @@ func GrabBanner(config *Config, target *GrabTarget) *Grab {
 		dial := makeDialer(config)
 		grabber := makeGrabber(config)
 		port := strconv.FormatUint(uint64(config.Port), 10)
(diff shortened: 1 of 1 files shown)


dadrian:
Does this work? Is `target.Domain` set if you don't pass an IP?

ajholland:
`target.Domain` would have to be set if someone specifies config.LookupDomain

I confirmed this works for the first couple issues described in #298

Issue #301, opened 2018-01-02
Make Zgrab use an Accept Header


launchramp:
Hi, I've been trying to get Zgrab to use an accept header when making requests as it only currently uses:
```
User-Agent: Mozilla/5.0 zgrab/0.x
Accept-Encoding: gzip
```
I was able to add the header using the http-req file, but when I run this it simplifies the output - I don't get the sha256 hash and the response headers aren't properly formatted in JSON anymore.

**http-req file**
```
GET / HTTP/1.1
Host: %s
Accept: */*
User-Agent: Mozilla/5.0 zgrab/0.x
Accept-Encoding: gzip
```

**using http-req file**
```
root@test:/opt/zgrab# echo "93.184.216.34" | ./zgrab --port 80 --data http-req --timeout 5 --http-max-size 5 --senders 5000 --http-max-redirects 2 | jq
Jan  2 03:27:03.029 [INFO] banner-grab: started grab at 2018-01-02T03:27:03Z
{
  "ip": "93.184.216.34",
  "timestamp": "2018-01-02T03:27:03Z",
  "data": {
    "read": "HTTP/1.1 404 Not Found\r\nContent-Type: text/html\r\nDate: Tue, 02 Jan 2018 03:27:03 GMT\r\nServer: ECS (oxr/839F)\r\nContent-Length: 345\r\n\r\n<?xml version=\"1.0\" encoding=\"iso-8859-1\"?>\n<!DOCTYPE html PUBLIC \"-//W3C//DTD XHTML 1.0 Transitional//EN\"\n         \"http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd\">\n<html xmlns=\"http://www.w3.org/1999/xhtml\" xml:lang=\"en\" lang=\"en\">\n\t<head>\n\t\t<title>404 - Not Found</title>\n\t</head>\n\t<body>\n\t\t<h1>404 - Not Found</h1>\n\t</body>\n</html>\n",
    "write": "GET / HTTP/1.1\r\nHost: 93.184.216.34\r\nAccept: */*\r\nUser-Agent: Mozilla/5.0 zgrab/0.x\r\nAccept-Encoding: gzip\r\n\r\n"
  }
}
Jan  2 03:27:03.263 [INFO] banner-grab: finished grab (1 success; 0 failure) at 2018-01-02T03:27:03Z
{
  "port": 80,
  "success_count": 1,
  "failure_count": 0,
  "total": 1,
[...]

PR #302, opened 2018-01-02
Add Accept HTTP header

justinbastress:
For `http` zgrab requests, add an `Accept: */*` header (addresses issue #301)

### DIFF
--- a/zlib/grabber.go
+++ b/zlib/grabber.go
@@ -295,8 +295,6 @@ func makeHTTPGrabber(config *Config, grabData *GrabData) func(string, string, st
 			fullURL = "http://" + urlHost + endpoint
 		}
 
-		var resp *http.Response
-
 		u, err := url.Parse(fullURL)
 		if err != nil {
 			return err
@@ -314,7 +312,10 @@ func makeHTTPGrabber(config *Config, grabData *GrabData) func(string, string, st
 			}
 			httpHost = hostWithoutPort
 		}
+
 		var req *http.Request
+		var resp *http.Response
+
 		switch config.HTTP.Method {
 		case "GET":
 			req, err = http.NewRequestWithHost("GET", fullURL, httpHost, nil)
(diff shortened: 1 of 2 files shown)


zlib/grabber.go
```diff
@@ -315,14 +313,21 @@ func makeHTTPGrabber(config *Config, grabData *GrabData) func(string, string, st
 			httpHost = hostWithoutPort
 		}
 
+		var req *http.Request
+		var resp *http.Response
+
 		switch config.HTTP.Method {
 		case "GET":
-			resp, err = client.GetWithHost(fullURL, httpHost)
+			req, err = http.NewRequestWithHost("GET", fullURL, httpHost, nil)
```
zakird: Can you explain why the call changes here?

justinbastress: Sure -- the Get/HeadWithHost does the request immediately, before we get a chance to add any headers. It would be nice if you could set default headers in the client, but I don't see an API for that. The docs for GetWithHost say "To make a request with custom headers, use NewRequest and Client.Do."

zakird: Got it! Thanks for the explanation!

zakird:
This looks good to me offhand. It might be worth running this on a sample of Alexa (e.g., 1000 random domains) and seeing if we're losing any results/data in the output, given that this a fairly core piece of Censys.

zakird:
Hey @ajholland would you be willing to take a quick glance at this one since you implemented much of this HTTP code?

justinbastress:
@zakird Ran the patched code and the code currently in master against the top 1000 sites 10 times; the relevant metric seemed to sites that always failed for a given version of the code.

For the old client, there were two site that failed each time: nike.com, where it gave a 403 as opposed to a 200, and playstation.com, where it gave up after 2 redirects (the new client got a 403 forbidden after the redirect to SSL).

For the new client, there were no sites that failed each time.

ajholland:
lgtm

Thanks for the ping @zakir

zakird:
🎉

PR #303, opened 2018-01-09
Skip validation

jcrowgey:
This feature implements the `InsecureSkipValidation` option proposed in Zmap/zcrypto#103 as `--tls-skip-validation`.


jcrowgey:
Note: CI is expected to fail until https://github.com/zmap/zcrypto/pull/103 lands.

justinbastress:
@jcrowgey I attempted to re-create https://github.com/zmap/zcrypto/pull/103 at https://github.com/zmap/zcrypto/pull/110, but with the crucial change that certificates that fail to parse still cause a failure regardless of whether the new flag is passed in. I originally thought that was the intention, but that is contradicted by the original docs which say that it doesn't even *parse* the certificate.

If that aligns with your original intent, then this can be merged after that.

If not --  then https://github.com/zmap/zcrypto/pull/110 needs to be fixed first.

jcrowgey:
No, what happened is that InsecureSkipValidation, as I wrote it, doesn't scale to Internet proportions as there are sites out there which will return malformed certificates.  If you don't parse these, then you don't know that they're malformed.  If you don't know that they're malformed, then you don't know that you won't be able to pull a server pubkey out of them, so you can't continue down the TLS handshake path.  I locally tested a version which parsed only the server leaf cert, and this worked, but it slowed things down enough that (for now) I prefer to scan as in my previous PR ---tls-certs-only, aborting the handshake after `ServerCertificates` and postprocessing the collected data offline.

If there's community interest in InsecureSkipValidation (where I parse only the first cert but skip validation), this is still quite a bit faster than the default implementaiton, but not as fast as certs-only, then I could submit the fix that i worked on locally.  Otherwise, if this PR was just for my own work, then we can just close it as I've moved on and decided I'm happy with --tls-certs-only.  It makes some noise in very verbose logs on the scanned servers, but that's, unfortunately, just part of the cost of doing business if you want to scan hundreds of millions of hosts relatively quickly.

PR #305, opened 2018-01-29
zgrab: log target address/port on panic

justinbastress:
If there is a `panic()` while scanning a target in GrabBanner, trap it and log the target to the standard logger at the ERROR level, before re-throwing the panic.

## How to Test

There is no place place that intentionally panics, so I have been testing by adding a panic() call to e.g. `Scan@modules/ftp.go`.

## Notes & Caveats

This arose from debugging https://ichnaea.eecs.umich.edu/admin/tasks/a47e2ae2-4f47-4ae4-86da-b8cc3a85a5c9 -- since the code panicked, the IP address did not get written to the output. We don't log the IP in non-panic cases since in those cases, the IP address should be available in the output.

## Issue Tracking

Re https://trello.com/c/CcL2qo60/1191-zgrab-https-scan-segfault, which was fixed by https://github.com/zmap/zcrypto/pull/107.

### DIFF
--- a/zlib/grabber.go
+++ b/zlib/grabber.go
@@ -24,7 +24,6 @@ import (
 	"io"
 	"net"
 	"net/url"
-	"runtime/debug"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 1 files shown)


zakird:
oh that's awesome! I didn't know you could do that in golang!

PR #311, opened 2018-03-05
Add vendored dependencies and Gopkg files

justinbastress:
All of this was auto-generated by running `dep init -v`; the only modifications were to the comments in the Gopkg.toml file, where I noted the date / commit where it was generated.


justinbastress:
It seemed to work with a spot check of a few dozen hosts; currently awaiting the results of a 1% scan HTTPS.

justinbastress:
before/after 1% scan looks good; out of ~500k hosts scanned, 2055 failed with the patch but not without, while 3363 failed without the patch but not with it, or about 0.5%. Iterating the process on the stubborn leftovers led to 814, then 40, then 5.

So, it looks like there is no change -- which makes sense, since the code should be identical.

PR #322, opened 2018-06-05
Include both request and response headers in knownHeaders

justinbastress:
This problem was identified in zgrab2 at https://github.com/zmap/zgrab2/issues/67.

Previously, most of the request-only known headers were ending up in the "unknown" list.

This must not be merged into master until schemas have been updated.


zakird (approved):

justinbastress:
Note -- despite the apparent removals in the diff, this is an append-only schema change (the apparent removals are a side effect of sorting the merged list -- comparing the items individually shows that no old values were removed or changed).

dadrian:
Closing this out, since we've moved development to ZGrab2.

PR #327, opened 2018-11-01
Catch nil header `exclude` map

justinbastress:
`exclude` can be nil in a few cases (e.g. explicitly in header.Write(), but also in some cases of chunkWriter.writeHeader).


dadrian:
@justinbastress you're more familiar with this than I am, but LGTM % the failing tests.

justinbastress:
This is not the correct solution. In this context, `exclude == nil` would just mean that exclude is the empty map, and so nothing should be excluded -- but this makes *everything* excluded.

But, it's not clear what *could* cause a SIGSEGV at address=0x0 at header.go:191...

PR #332, opened 2020-05-11
Add Go modules

dadrian:
Remove Gopkg in favor of Go modules for vendoring. This should make it easier for anyone depending on this (legacy) project to build. If Gopkg is required, users can just point to a previous commit.

```console
$ go mod init github.com/zmap/zgrab
go: creating new go.mod: module github.com/zmap/zgrab
go: copying requirements from Gopkg.lock
$ go mod vendor
$ rm -f Gopkg.*
```


dadrian:
@justinbastress I think that go modules and Gopkg resolved something different in the mod-to-sum stage with regard to transitive dependencies. Any ideas what's going on / if we're OK?

